PR #4, opened 2023-03-13
Add magnet coil modeling

connoramoreno:
Add modeling functionality for magnet coils based on user-supplied point-locus data

### DIFF
--- a/magnet_coils.py
+++ b/magnet_coils.py
@@ -1,63 +1,67 @@
 import cubit
 import numpy as np
+import logging
 
 
-def orient_rectangle(path_origin, surf_id, norm, rot_axis, rot_ang_norm):
-    """Orients rectangular cross-section in the normal plane such that its thickness direction faces the origin.
+# Get logger defined in parametric_stellarator.py
+logger = logging.getLogger('log')
+
+
+def orient_rectangle(path_origin, surf_id, t_vec, norm, rot_axis, rot_ang_norm):
+    """Orients rectangular cross-section in the normal plane such that its
+    thickness direction faces the origin.
     
     Arguments:
         path_origin (int): index of initial point in filament path.
         surf_id (int): index of cross-section surface.
+        t_vec (list of float): cross-section thickness vector.
         norm (list of float): cross-section normal vector.
         rot_axis (list of float): axis about which to rotate the cross-section.
         rot_ang_norm (float): angle by which cross-section was rotated to align
             its normal with the initial point tangent (deg).
     """
-    # Extract initial path point
-    pos = cubit.vertex(path_origin).coordinates()
+    # Determine orientation of thickness vector after cross-section was
+    # oriented along filament origin tangent
 
-    # Cross-section inititally populated with thickness vector
-    # oriented along x axis
-    t_vec = np.array([1, 0, 0])
     # Compute part of thickness vector parallel to rotation axis
-    t_vec_par = (np.inner(t_vec, rot_axis)/np.inner(rot_axis,
-        rot_axis))*rot_axis
+    t_vec_par = np.inner(t_vec, rot_axis)*rot_axis
+    t_vec_par = t_vec_par/np.linalg.norm(t_vec_par)
     # Compute part of thickness vector orthogonal to rotation axis
     t_vec_perp = t_vec - t_vec_par
-    # Compute magnitude of orthogonal part of thickness vector
-    t_vec_perp_mag = np.linalg.norm(t_vec_perp)
+    t_vec_perp = t_vec_perp/np.linalg.norm(t_vec_perp)
 
     # Compute vector othogonal to both rotation axis and orthogonal
     # part of thickness vector
     orth = np.cross(rot_axis, t_vec_perp)
-    # Compute magnitude of orthgonal vector
-    orth_mag = np.linalg.norm(orth)
+    orth = orth/np.linalg.norm(orth)
 
     # Convert angle from degress to radians
     rot_ang_norm = np.deg2rad(rot_ang_norm)
     
     # Determine part of rotated vector parallel to original
-    rot_par = np.cos(rot_ang_norm)/t_vec_perp_mag
+    rot_par = np.cos(rot_ang_norm)
     # Determine part of rotated vector orthogonal to original
-    rot_perp = np.sin(rot_ang_norm)/orth_mag
+    rot_perp = np.sin(rot_ang_norm)
 
     # Compute orthogonal part of thickness vector after rotation
-    t_vec_perp_rot = t_vec_perp_mag*(rot_par*t_vec_perp
-        + rot_perp*orth)
+    t_vec_perp_rot = rot_par*t_vec_perp + rot_perp*orth
     # Compute thickness vector after rotation
     t_vec_rot = t_vec_perp_rot + t_vec_par
-    # Compute magnitude of thickness vector
-    t_vec_rot_mag = np.linalg.norm(t_vec_rot)
+    t_vec_rot = t_vec_rot/np.linalg.norm(t_vec_rot)
+
+    # Orient cross-section in its plane such that it faces the global origin
+
+    # Extract initial path point
+    pos = cubit.vertex(path_origin).coordinates()
 
     # Project position vector onto cross-section
     pos_proj = pos - (np.inner(pos, norm)/np.inner(pos, pos))*norm
-    # Compute magnitude of projection of position vector
-    pos_proj_mag = np.linalg.norm(pos_proj)
+    pos_proj = pos_proj/np.linalg.norm(pos_proj)
 
     # Compute angle by which to rotate cross-section such that it
     # faces the origin
     dot_prod = np.inner(pos_proj, t_vec_rot)
-    rot_ang_orig = np.arccos(dot_prod/t_vec_rot_mag/pos_proj_mag)
+    rot_ang_orig = np.arccos(dot_prod)
 
     # Convert angle from radians to degrees
     rot_ang_orig = np.rad2deg(rot_ang_orig)
@@ -65,24 +69,84 @@ def orient_rectangle(path_origin, surf_id, norm, rot_axis, rot_ang_norm):
     # Re-orient rotated cross-section such that thickness vector
     # faces origin
     cubit.cmd(
-        f'rotate Surface {surf_id} angle {rot_ang_orig} about  origin 0 0 0 '
-        f'direction {" ".join(str(i) for i in norm)} include_merged'
+        f'rotate Surface {surf_id} angle {rot_ang_orig} about  origin 0 0 0 ' +
+        'direction ' + ' '.join(str(i) for i in norm)
     )
 
 
-def create_magnets(
-    filaments, shape, radius = None, width = None, thickness = None):
+def create_magnets(filaments, cross_section):
     """Creates magnet coil solids.
     
     Arguments:
         filaments (list of list of list of float): list filament coordinates.
             Each filament is a list of coordinates.
-        shape (str): shape of cross-section (must be 'circle' or 'rectangle').
-        radius (float): radius of circular cross-section (defaults to None).
-        width (float): width of rectangular cross-seciton (defaults to None).
-        thickness (float): thickness of rectangular cross-section (defaults to
-            None).
+        cross_section (list or tuple of str, float, float): coil cross-section
+            definition. Note that the cross-section shape must be either a
+            circle or rectangle.
+            For a circular cross-section, the list format is
+            ['circle' (str), radius (float, cm)]
+            For a rectangular cross-section, the list format is
+            ['rectangle' (str), width (float, cm), thickness (float, cm)]
+
+    Returns:
+        vol_id (list of int): indices for magnet volumes.
     """
+    # Extract coil cross-section shape
+    shape = cross_section[0]
+    # Cross-section inititally populated with thickness vector
+    # oriented along x axis
+    t_vec = np.array([1, 0, 0])
+    # Conditionally extract parameters for circular cross-section
+    if shape == 'circle':
+        if len(cross_section) != 2:
+            logger.debug(
+                'Format of list defining circular cross-section must be\n'
+                '[\'circle\' (str), radius (float, cm)]'
+            )
+            raise ValueError(
+                'Format of list defining circular cross-section must be\n'
+                '[\'circle\' (str), radius (float, cm)]'
+            )
+        radius = cross_section[1]
+    # Conditinally extract parameters for rectangular cross-section
+    elif shape == 'rectangle':
+        if len(cross_section) != 3:
+            logger.debug(
+                'Format of list defining rectangular cross-section must be\n'
+                '[\'rectangle\' (str), width (float, cm), thickness '
+                '(float, cm)]'
+            )
+            raise ValueError(
+                'Format of list defining rectangular cross-section must be\n'
+                '[\'rectangle\' (str), width (float, cm), thickness '
+                '(float, cm)]'
+            )
+        width = cross_section[1]
+        thickness = cross_section[2]
+    else:
+        logger.debug(
+            'Magnet cross-section must be either a circle or rectangle. The '
+            'first entry of the list defining the cross-section must be the '
+            'shape, with the following entries defining the shape parameters.\n'
+            '\n'
+            'For a circular cross-section, the list format is\n'
+            '[\'circle\' (str), radius (float, cm)]\n'
+            '\n'
+            'For a rectangular cross-section, the list format is\n'
+            '[\'rectangle\' (str), width (float, cm), thickness (float, cm)]'
+        )
+        raise ValueError(
+            'Magnet cross-section must be either a circle or rectangle. The '
+            'first entry of the list defining the cross-section must be the '
+            'shape, with the following entries defining the shape parameters.\n'
+            '\n'
+            'For a circular cross-section, the list format is\n'
+            '[\'circle\' (str), radius (float, cm)]\n'
+            '\n'
+            'For a rectangular cross-section, the list format is\n'
+            '[\'rectangle\' (str), width (float, cm), thickness (float, cm)]'
+        )
+    
     # Create cross-section for sweep
     if shape == 'circle':
         cubit.cmd(
@@ -94,6 +158,12 @@ def create_magnets(
         )
     # Store cross-section index
     cs_id = cubit.get_last_id("surface")
+    # Cross-section initially populated with normal oriented along z
+    # axis
+    cs_axis = np.array([0, 0, 1])
+
+    # Initialize volume index storage list
+    vol_id = []
 
     # Initialize path list
     path = []
@@ -102,9 +172,7 @@ def create_magnets(
     for filament in filaments:
         
         # Create vertices in filament path
-        for coords in filament:
-            # Extract Cartesian coordinates in filament
-            x, y, z = coords
+        for x, y, z in filament:
             # Create vertex in path
             cubit.cmd(f'create vertex {x} {y} {z}')
             # Append vertex to path list
@@ -115,16 +183,11 @@ def create_magnets(
         
         # Create spline for path
         cubit.cmd(
-            f'create curve spline location vertex '
-            f'{" ".join(str(i) for i in path)}'
+            f'create curve spline location vertex ' +
+            ' '.join(str(i) for i in path)
         )
         # Store curve index
         curve_id = cubit.get_last_id("curve")
-        
-        # Copy cross-section for sweep
-        cubit.cmd(f'surface {cs_id} copy')
-        # Store surface index
-        surf_id = cubit.get_last_id("surface")
 
         # Define new surface normal vector as that pointing between path
         # points adjacent to initial point
@@ -132,50 +195,53 @@ def create_magnets(
         # Extract path points adjacent to initial point
         next_pt = np.array(cubit.vertex(path[1]).coordinates())
         last_pt = np.array(cubit.vertex(path[-2]).coordinates())
-        # Compute surface normal
-        norm = np.subtract(next_pt, last_pt)
-        # Compute magnitude of surface normal
-        norm_mag = np.linalg.norm(norm)
+        # Compute tangent at path origin (direction in which to align surface
+        # normal)
+        tang = np.subtract(next_pt, last_pt)
+        tang = tang/np.linalg.norm(tang)
         
         # Define axis and angle of rotation to orient cross-section along
         # defined normal
 
-        # Cross-section initially populated with normal oriented along z
-        # axis
-        z_axis = np.array([0, 0, 1])
         # Define axis of rotation as orthogonal to both z axis and surface
         # normal
-        rot_axis = np.cross(z_axis, norm)
+        rot_axis = np.cross(cs_axis, tang)
+        rot_axis = rot_axis/np.linalg.norm(rot_axis)
         # Compute angle by which to rotate cross-section to orient along
         # defined surface normal
-        dot_prod = np.inner(z_axis, norm)
-        rot_ang_norm = np.arccos(dot_prod/norm_mag)
+        dot_prod = np.inner(cs_axis, tang)
+        rot_ang_norm = np.arccos(dot_prod)
 
         # Covert angle from radians to degrees
         rot_ang_norm = np.rad2deg(rot_ang_norm)
 
+        # Copy cross-section for sweep
+        cubit.cmd(f'surface {cs_id} copy')
+        # Store surface index
+        surf_id = cubit.get_last_id("surface")
+        
         # Orient cross-section along defined normal
         cubit.cmd(
-            f'rotate Surface {surf_id} angle {rot_ang_norm} about '
-            f'origin 0 0 0 direction {" ".join(str(i) for i in rot_axis)} '
-            f'include_merged'
+            f'rotate Surface {surf_id} angle {rot_ang_norm} about ' +
+            'origin 0 0 0 direction ' + ' '.join(str(i) for i in rot_axis)
         )
 
         # Conditionally orients rectangular cross-section
         if shape == 'rectangle':
-            orient_rectangle(path[0], surf_id, norm, rot_axis, rot_ang_norm)
+            orient_rectangle(
+                path[0], surf_id, t_vec, tang, rot_axis, rot_ang_norm
+            )
             
         # Move cross-section to initial path point
-        cubit.cmd(
-            f'move Surface {surf_id} location vertex {path[0]} '
-            f'include_merged'
-        )
+        cubit.cmd(f'move Surface {surf_id} location vertex {path[0]}')
 
         # Sweep cross-section to create magnet coil
         cubit.cmd(
             f'sweep surface {surf_id} along curve {curve_id} '
             f'individual'
         )
+        # Store volume index
+        vol_id.append(cubit.get_last_id("volume"))
         # Delete extraneous curves and vertices
         cubit.cmd(f'delete curve {curve_id}')
         cubit.cmd('delete vertex all')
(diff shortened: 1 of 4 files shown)


magnet_coils.py
```diff
@@ -0,0 +1,247 @@
...
+
+        # Extract coordinates for vertex in path
+        x = float(columns[0])*100
+        y = float(columns[1])*100
+        z = float(columns[2])*100
+        # Extract coil current
+        s = float(columns[3])
+        
+        # Create vertex in path
+        cubit.cmd('create vertex ' + str(x) + ' ' + str(y) + ' ' + str(z))
```
biegelk, line 90: f-strings are a very convenient and more compact alternative to concatenation: https://docs.python.org/3/tutorial/inputoutput.html

parametric_stellarator.py
```diff
@@ -470,9 +528,24 @@ def parametric_stellarator(
...
+    if include_magnets == True:
+        # Extract coil data
+        file = magnets['file']
+        cross_section = magnets['cross_section']
+        start = magnets['start']
+        stop = magnets['stop']
+        magnet_name = magnets['name']
+        magnet_h5m_tag = magnets['h5m_tag']
+        # Generate magnets
+        magnet_coils.magnet_coils(file, cross_section, start, stop, magnet_name)
```
gonuke, line 541: Why not just pass in `magnets` as an argument rather than extracting all the values here into individual variables?

parametric_stellarator.py
```diff
@@ -470,9 +528,24 @@ def parametric_stellarator(
     if include_graveyard == True:
         components = graveyard(vmec, offset, components)
 
+    # Conditionally build magnet coils
+    if include_magnets == True:
```
gonuke, line 532:
```suggestion
    if include_magnets:
```

parametric_stellarator.py
```diff
@@ -353,9 +379,10 @@ def stellarator_torus(
 
 def parametric_stellarator(
     plas_eq, num_periods, radial_build, gen_periods, num_phi = 60,
-    num_theta = 100, exclude = [], step_export = True, h5m_export = None,
-    plas_h5m_tag = None, include_graveyard = False, facet_tol = None,
-    len_tol = None, norm_tol = None, min_mesh_size = 5.0, max_mesh_size = 20.0,
+    num_theta = 100, exclude = [], plas_h5m_tag = None,
```
gonuke, line 382: We should discuss strategies to manage this ballooning list of arguments, possibly with 1 or more dictionaries

gonuke (changes requested):
In addition to these comments, we should look at some high level design questions for maintainability and readability.

magnet_coils.py
connoramoreno: This is really cool, thanks!

magnet_coils.py
```diff
@@ -0,0 +1,247 @@
...
+            'For a circular cross-section, the list format is\n'
+            '[\'circle\' (str), radius (float, cm)]\n'
+            '\n'
+            'For a rectangular cross-section, the list format is\n'
+            '[\'rectangle\' (str), width (float, cm), thickness (float, cm)]'
+        )
+    
+    # Load magnet coil data
+    file_obj = open(file, 'r')
+    data = file_obj.readlines()[start:stop]
```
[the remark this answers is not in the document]
connoramoreno: If `stop == None`, the file will be read to its last line. `stop` is just an optional variable if the user wants to terminate reading at a specified line.

magnet_coils.py
```diff
@@ -0,0 +1,247 @@
...
+                # Compute magnitude of projection of position vector
+                pos_proj_mag = np.linalg.norm(pos_proj)
+
+                # Compute angle by which to rotate cross-section such that it
+                # faces the origin
+                dot_prod = np.inner(pos_proj, t_vec_rot)
+                rot_ang_orig = np.arccos(dot_prod/t_vec_rot_mag/pos_proj_mag)
+
+                # Convert angle from radians to degrees
+                rot_ang_orig = np.rad2deg(rot_ang_orig)
```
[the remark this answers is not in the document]
connoramoreno: The `cubit_export` function and `magnet_coils` script provide different functionality, so my thought here was to initialize Cubit separately in case one would like to use one and not the other.

parametric_stellarator.py
```diff
@@ -19,15 +22,24 @@ def cubit_export(solids, h5m_tags, facet_tol, len_tol, norm_tol):
...
-    # Initialize Cubit
-    cubit.init([''])
+    # If not already initialized by generating magnets, initialize Cubit and
+    # volume index
+    if include_magnets == False:
+        cubit.init([''])
+        vol_id = 1
+    # If magnets generated, retrieve index of last volume created
+    else:
+        vol_id = cubit.get_last_id("volume")
```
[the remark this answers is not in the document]
connoramoreno: In this version of the code, the H5M tag for the magnets is input via the `magnets` dictionary, and this H5M tag is not used in the `magnet_coils` script. I can have `magnet_h5m_tag` passed separately though so that the arguments don't need to be parsed before calling `magnet_coils`

parametric_stellarator.py
connoramoreno: Yes, it is getting a bit cumbersome. One idea is to have different aspects of the model (e.g., export, magnets, source mesh later on, etc.) passed as their own dictionaries

magnet_coils.py
```diff
@@ -0,0 +1,247 @@
...
+        x = float(columns[0])*100
+        y = float(columns[1])*100
+        z = float(columns[2])*100
+        # Extract coil current
+        s = float(columns[3])
+        
+        # Create vertex in path
+        cubit.cmd('create vertex ' + str(x) + ' ' + str(y) + ' ' + str(z))
+        # Update vertex index
+        vert_id += 1
```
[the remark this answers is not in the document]
gonuke: I think it will make things more portable, especially as we consider other changes for separation of concerns

parametric_stellarator.py
connoramoreno: In this version of the code, the H5M tag for the magnets is input via the `magnets` dictionary, and this H5M tag is not used in the `magnet_coils` script. I can have `magnet_h5m_tag` passed separately though so that the arguments don't need to be parsed before calling `magnet_coils`

connoramoreno:
It seems my comments somehow (perhaps my own error) were made on the conversation above that which I was actually trying to comment on. I copied and pasted the comments to the appropriate conversations.

parametric_stellarator.py
```diff
@@ -497,12 +597,12 @@ def parametric_stellarator(
     for name, layer_data in full_radial_build.items():
         
         # Notify which component is being generated
-        print('Generating ' + str(name) + '...')
+        print(f'Generating {name}...')
```
gonuke, line 600: explore using a `logger` to better control when output occurs

parametric_stellarator.py
```diff
@@ -519,33 +619,22 @@ def parametric_stellarator(
...
             components[name]['solid'] = torus
             components[name]['h5m_tag'] = layer_data['h5m_tag']
 
     # Conditionally build graveyard volume
-    if include_graveyard == True:
+    if export_dict['graveyard']:
         components = graveyard(vmec, offset, components)
 
+    # Conditionally initialize Cubit
+    if export_dict['h5m_export'] == 'Cubit':
```
gonuke, line 632: Also if magnets are defined

parametric_stellarator.py
```diff
@@ -1,45 +1,107 @@
...
 
     # Imprint and merge all volumes
     cubit.cmd('imprint volume all')
     cubit.cmd('merge volume all')
 
+    # Conditionally assign magnet material group
+    if magnets is not None:
+        magnet_h5m_tag = magnets['h5m_tag']
+        cubit.cmd(
+            f'group "mat:{magnet_h5m_tag}" add volume {" ".join(str(i) for i in np.linspace(1, vol_id, num = vol_id))}'
```
gonuke, line 97: I don't think we should rely on a hard-coded volume IDs starting at 1.  Makes this very inflexible for future changes.

parametric_stellarator.py
```diff
@@ -1,45 +1,107 @@
...
+        )
+        # Increment volume index for consecutive material tagging
+        vol_id += 1
+    
     # Assign material groups
-    for i, tag in enumerate(h5m_tags, start = 1):
-        cubit.cmd('group "mat:' + tag + '" add volume ' + str(i))
+    for i, tag in enumerate(h5m_tags, start = vol_id):
```
gonuke, line 103: I'm concerned about assuming that `h5m_tags` and the list of `step` files are always in the right order.  We should discuss this data structure.

parametric_stellarator.py
```diff
@@ -414,11 +556,27 @@ def parametric_stellarator(
         )
 
     # Check that h5m_export has an appropriate value
-    if h5m_export != None and h5m_export != 'Cubit' and h5m_export != 'Gmsh':
+    if (export['h5m_export'] != None and
+        export['h5m_export'] != 'Cubit' and
+        export['h5m_export'] != 'Gmsh'):
```
gonuke, line 561:
```suggestion
    if export['h5m_export'] not in ['Cubit', 'Gmsh']:
```

parametric_stellarator.py
```diff
@@ -414,11 +556,27 @@ def parametric_stellarator(
...
         raise ValueError(
             'h5m_export must be None or have a string value of \'Cubit\' or '
             '\'Gmsh\''
         )
+
+    # Check that Cubit export has STEP files to use
+    if export['step_export'] == False and export['h5m_export'] == 'Cubit':
```
gonuke, line 568:
```suggestion
    if export['h5m_export'] == 'Cubit' and not export['step_export']:
```

gonuke (changes requested):
Most of my comments are aimed at long-term readability and maintainability.

parametric_stellarator.py
```diff
@@ -1,45 +1,107 @@
...
 
-    # Initialize Cubit
-    cubit.init([''])
+    # Initialize volume index
+    vol_id = cubit.get_last_id("volume")
 
     # Import solids
     for solid in solids:
-        cubit.cmd(impt_cmd + solid + '.step" heal')
+        cubit.cmd(f'{impt_cmd}{solid}.step" heal')
```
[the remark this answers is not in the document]
connoramoreno: Would it be better just to concatenate the strings?

magnet_coils.py
```diff
@@ -0,0 +1,317 @@
...
+    """Creates magnet coil solids.
+    
+    Arguments:
+        filaments (list of list of list of float): list filament coordinates.
+            Each filament is a list of coordinates.
+        shape (str): shape of cross-section (must be 'circle' or 'rectangle').
+        radius (float): radius of circular cross-section (defaults to None).
+        width (float): width of rectangular cross-seciton (defaults to None).
+        thickness (float): thickness of rectangular cross-section (defaults to
+            None).
```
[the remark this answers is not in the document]
connoramoreno: I've changed it so that `cross_section` gets passed to and parsed inside the `create_magnets` method. An alternate way of doing things would be to pass the `magnets` dictionary instead and extract `cross_section` from there, and do the same thing for the `extract_filaments` method. Let me know what you think!

magnet_coils.py
```diff
@@ -0,0 +1,317 @@
...
+        surf_id = cubit.get_last_id("surface")
+
+        # Define new surface normal vector as that pointing between path
+        # points adjacent to initial point
+
+        # Extract path points adjacent to initial point
+        next_pt = np.array(cubit.vertex(path[1]).coordinates())
+        last_pt = np.array(cubit.vertex(path[-2]).coordinates())
+        # Compute surface normal
+        norm = np.subtract(next_pt, last_pt)
```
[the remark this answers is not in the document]
connoramoreno: How about `tang`? As in the tangent at the path (filament) origin?

magnet_coils.py
```diff
@@ -0,0 +1,317 @@
...
+        )
+
+        # Sweep cross-section to create magnet coil
+        cubit.cmd(
+            f'sweep surface {surf_id} along curve {curve_id} '
+            f'individual'
+        )
+        # Delete extraneous curves and vertices
+        cubit.cmd(f'delete curve {curve_id}')
+        cubit.cmd('delete vertex all')
```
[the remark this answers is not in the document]
connoramoreno: delete vertex all will only delete free vertices, i.e., the vertices added in the path (except those on the swept cross-section)

magnet_coils.py
```diff
@@ -0,0 +1,317 @@
...
+            break
+
+        # Extract coordinates for vertex in path
+        x = float(columns[0])*100
+        y = float(columns[1])*100
+        z = float(columns[2])*100
+        # Extract coil current
+        s = float(columns[3])
+
+        # Coil current of zero signals end of filament
```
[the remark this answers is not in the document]
connoramoreno: This gets reset before adding the newest coordinates so that the final filament point is not appended to the list. To create a continuous path in Cubit, the initial and final points in the curve must be the same vertex. The documentation I wrote is actually incorrect, the issue is not that the coordinates must be equal, it's that the vertex ID must be equal. So, in the `create_magnets` method, I ensure the final curve vertex for each filament is the same as the initial. I will update the documentation to reflect this.

magnet_coils.py
```diff
@@ -0,0 +1,317 @@
...
+            f'create curve spline location vertex '
+            f'{" ".join(str(i) for i in path)}'
+        )
+        # Store curve index
+        curve_id = cubit.get_last_id("curve")
+        
+        # Copy cross-section for sweep
+        cubit.cmd(f'surface {cs_id} copy')
+        # Store surface index
+        surf_id = cubit.get_last_id("surface")
```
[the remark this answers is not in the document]
connoramoreno: The vector math in `create_magnets` is similar to that in `orient_rectangle` but they are not quite the same. The blocks that are similar are about 2 lines long each so I'm not sure it's worth creating an additional function

magnet_coils.py
```diff
@@ -0,0 +1,350 @@
...
+            }
+            For the list defining the coil cross-section, the cross-section
+            shape must be either a circle or rectangle. For a circular
+            cross-section, the list format is
+            ['circle' (str), radius (float, cm)]
+            For a rectangular cross-section, the list format is
+            ['rectangle' (str), width (float, cm), thickness (float, cm)]
+
+    Returns:
+        vol_id (list of int): indices for magnet volumes.
```
gonuke, line 337: it might be more readable if this was plural `vol_ids`

magnet_coils.py
```diff
@@ -0,0 +1,350 @@
...
+            'Magnet cross-section must be either a circle or rectangle. The '
+            'first entry of the list defining the cross-section must be the '
+            'shape, with the following entries defining the shape parameters.\n'
+            '\n'
+            'For a circular cross-section, the list format is\n'
+            '[\'circle\' (str), radius (float, cm)]\n'
+            '\n'
+            'For a rectangular cross-section, the list format is\n'
+            '[\'rectangle\' (str), width (float, cm), thickness (float, cm)]'
+        )
```
gonuke, line 148: This might make a good function, especially if it just returned a string, `shape_str`, that was either:
* `f'{shape} radius {radius}'` or 
* `f'{shape} width {thickness} height {width}'`

so you could then just call

`cubit.cmd('create surface ' + shape_str + ' zplane')`


(Note: `t_vec` should be defined outside of this function, either before or after...)

magnet_coils.py
```diff
@@ -0,0 +1,350 @@
...
+        # Define new surface normal vector as that pointing between path
+        # points adjacent to initial point
+
+        # Extract path points adjacent to initial point
+        next_pt = np.array(cubit.vertex(path[1]).coordinates())
+        last_pt = np.array(cubit.vertex(path[-2]).coordinates())
+        # Compute tangent at path origin (direction in which to align surface
+        # normal)
+        tang = np.subtract(next_pt, last_pt)
+        tang = tang/np.linalg.norm(tang)
```
gonuke, line 201: What if we made a small function called:
```
def unit_vector(vec):
    return vec/np.linalg.norm(vec)
```
and then each of these two line sequences could become like:
`tang = unit_vector(np.subtract(next_pt, last_pt))`

magnet_coils.py
```diff
@@ -0,0 +1,350 @@
...
+        # defined normal
+
+        # Define axis of rotation as orthogonal to both z axis and surface
+        # normal
+        rot_axis = np.cross(cs_axis, tang)
+        rot_axis = rot_axis/np.linalg.norm(rot_axis)
+        # Compute angle by which to rotate cross-section to orient along
+        # defined surface normal
+        dot_prod = np.inner(cs_axis, tang)
+        rot_ang_norm = np.arccos(dot_prod)
```
gonuke, line 213: I think it's probably readable to combine these:
```suggestion
        rot_ang_norm = np.arccos(np.inner(cs_axis, tang))
```

magnet_coils.py
```diff
@@ -0,0 +1,350 @@
...
+    pos = cubit.vertex(path_origin).coordinates()
+
+    # Project position vector onto cross-section
+    pos_proj = pos - (np.inner(pos, norm)/np.inner(pos, pos))*norm
+    pos_proj = pos_proj/np.linalg.norm(pos_proj)
+
+    # Compute angle by which to rotate cross-section such that it
+    # faces the origin
+    dot_prod = np.inner(pos_proj, t_vec_rot)
+    rot_ang_orig = np.arccos(dot_prod)
```
gonuke, line 64: I think it's reasonable to combine these:
```suggestion
    rot_ang_orig = np.arccos(np.inner(pos_proj, t_vec_rot))
```

magnet_coils.py
```diff
@@ -0,0 +1,350 @@
...
+
+    # Convert angle from radians to degrees
+    rot_ang_orig = np.rad2deg(rot_ang_orig)
+    
+    # Re-orient rotated cross-section such that thickness vector
+    # faces origin
+    cubit.cmd(
+        f'rotate Surface {surf_id} angle {rot_ang_orig} about  origin 0 0 0 ' +
```
gonuke, line 72: Looking to streamline... 

Can we put the `rad2deg` conversion inline here?
```suggestion
        f'rotate Surface {surf_id} angle {np.rad2deg(rot_ang_orig)} about  origin 0 0 0 ' +
```

magnet_coils.py
```diff
@@ -0,0 +1,350 @@
...
+    t_vec_rot = t_vec_rot/np.linalg.norm(t_vec_rot)
+
+    # Orient cross-section in its plane such that it faces the global origin
+
+    # Extract initial path point
+    pos = cubit.vertex(path_origin).coordinates()
+
+    # Project position vector onto cross-section
+    pos_proj = pos - (np.inner(pos, norm)/np.inner(pos, pos))*norm
+    pos_proj = pos_proj/np.linalg.norm(pos_proj)
```
gonuke, line 59: If we have a `unit_vector` function this could become:
```suggestion
    pos_proj = unit_vector(pos - np.inner(unit_vector(pos),norm) * norm)
```

magnet_coils.py
```diff
@@ -0,0 +1,350 @@
...
+        # normal
+        rot_axis = np.cross(cs_axis, tang)
+        rot_axis = rot_axis/np.linalg.norm(rot_axis)
+        # Compute angle by which to rotate cross-section to orient along
+        # defined surface normal
+        dot_prod = np.inner(cs_axis, tang)
+        rot_ang_norm = np.arccos(dot_prod)
+
+        # Covert angle from radians to degrees
+        rot_ang_norm = np.rad2deg(rot_ang_norm)
```
gonuke, line 216: You convert this to degrees here to use in the cubit cmd below, and then pass it into the `orient_rectangle` function and convert it back to radians.  Maybe we can streamline this somewhat by doing the conversion to degrees inline in the cubit cmd below and otherwise keep it as radians.

gonuke (changes requested):
I've made a few more comments, many along the same lines, and some trying to simplify the vector math.

I think I also led you astray with the combination of logging and exception raising - we can discuss quickly if you like.

parametric_stellarator.py
```diff
@@ -1,45 +1,126 @@
...
+# Define and set logging format
+format = logging.Formatter(
+    fmt = '%(asctime)s: %(message)s',
+    datefmt = '%H:%M:%S'
+)
+s_handler.setFormatter(format)
+f_handler.setFormatter(format)
+# Add handlers to logger
+logger.addHandler(s_handler)
+logger.addHandler(f_handler)
```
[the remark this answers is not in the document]
connoramoreno: Why's that? So it can potentially be borrowed by other modules/scripts?

magnet_coils.py
connoramoreno: `np.inner(pos,pos) = magnitude(pos)**2`, so unless I'm missing something, I'm not sure this change would produce the same result

gonuke: Aaah - you're right, but now I wonder about why this is correct?  It seems like an unusual vector operation?

connoramoreno: The equation projects the position vector `pos` onto the cross-section plane (after the cross-section is oriented such that it's normal is aligned with the filament tangent). Essentially what it is doing is taking `pos` and subtracting from it the projection of `pos` onto `norm` (the cross-section normal, also the filament tangent), which is the component of `pos` that is normal to the cross-section. Thus, we are left with the component of `pos` that is parallel to the cross-section

parametric_stellarator.py
```diff
@@ -380,45 +507,100 @@ def parametric_stellarator(
...
+            'stellarator geometry')
         raise ValueError(
             'Number of requested periods to generate exceeds number in '
             'stellarator geometry'
         )
 
     # Check that h5m_export has an appropriate value
-    if h5m_export != None and h5m_export != 'Cubit' and h5m_export != 'Gmsh':
+    if export['h5m_export'] not in [None, 'Cubit', 'Gmsh']:
+        logger.debug(
```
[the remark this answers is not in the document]
connoramoreno: What I ended up doing instead is something of the form
```
# Check that h5m_export has an appropriate value
    if export['h5m_export'] not in [None, 'Cubit', 'Gmsh']:
        try:
            raise ValueError(
                'h5m_export must be None or have a string value of \'Cubit\' '
                'or \'Gmsh\''
            )
        except:
            logger.error('Exception occured', exc_info = True)
            sys.exit()
```
Do you think this works well?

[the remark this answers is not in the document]
pshriwise: I think that's always going to result in calling `logger.error`. Raising an exception inside the `try:` block with a naked `except:` will capture all raise exceptions/errors (including the one in the `try:` block) and execute the code in the `except:` block.

What I was thinking is more like an additional method that you can throw an error message at and it will either log the message, raise an exception, or both as needed.

```python
def error(self, message, exception=None):
    # might want to get a default logger here instead as discussed in the software meeting
    if self.logger:
        logger.error(message)
        sys.exit()
    if exception is not None:
        raise exception(message)
```

Or something close to that. Just something that keeps you from having to create the message more than once since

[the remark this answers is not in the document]
gonuke: Yeah! My bad that we didn't discuss this today. It's slightly related to the `logger` discussion.  I think your code should only raise exceptions.  The code that calls it should decide how to handle this exceptions, either:
• ignore it so that the exception continues up the stack
• catch it and put the message onto the log
• catch it, put the message on the log and continue to raise it

Logging + raising an exception at the same time seems to be defeating the purpose of exceptions.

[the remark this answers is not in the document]
connoramoreno: My goal was to have logger.error get called every time. I formatted it this way so that when the error occurs, instead of just raising a ValueError, the logger would also capture it along with the traceback, and from what I found online, this was the only way to do so.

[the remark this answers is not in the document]
gonuke: I think this gets into the philosophy of exceptions a little.  Fundamentally, exceptions are a way of raising an error/warning state up the call stack to be handled by the code that called it.  Because the philosophy is for the calling code to handle the exception, it should be that calling code that decides whether or not the message ends up in the log.  Forcing every exception message to also end up in the log kind of undermines that philosophy because it means that the log will get that message regardless of how the calling code wants to handle the exception.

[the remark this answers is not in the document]
connoramoreno: Based on an internet perusal, it seems like assert() would fit nicely here. Although this is different from Patrick's suggestion, how do we feel about handling the exceptions like
```
try:
    assert(
        export['h5m_export'] not in [None, 'Cubit', 'Gmsh']
    )
except AssertionError as err:
    logger.exception(
        'h5m_export must be None or have a string value of \'Cubit\' or \'Gmsh\''
    )
    raise err
```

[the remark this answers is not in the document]
pshriwise: Good point, @gonuke. Probably don't want all of these exception messages polluting the log if they are ignored further up the stack. I'm not sure how you would capture the stack output in the logger only if the exception is raised. Never attempted that before, but maybe we could look into it at the next software meeting.

@connoramoreno I understand you want to get the stack trace to the log here, but this approach seems a little convoluted. Any reason not to use the `traceback` module? Is there other info you're looking to capture in the log?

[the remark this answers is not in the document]
pshriwise: >  Mainly, I just figured that if we're going to have a log, it could be useful to capture the specific error and traceback. 

I think it's a good idea, but it might depend on your definition of a log and what its purpose is. Goes back to our discussion on verbosity levels maybe.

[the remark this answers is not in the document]
connoramoreno: I'd like to resolve this sooner rather than later but maybe it would be best to discuss this at the software meeting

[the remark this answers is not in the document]
pshriwise: I definitely don't want to slow progress here, so don't consider any of my comments to be blocking plz. I'll leave the decision on logging & exceptions to you and @gonuke. Hopefully the `traceback` module is helpful in simplifying things for you!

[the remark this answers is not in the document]
biegelk: Using your custom `logger` setup, this StackOverflow answer seemed to work: https://stackoverflow.com/a/37992988

```
try:
    <something that doesn't work>
except:
    logger.error("h5m_export must be None, 'Cubit', or 'Cmsh'", exc_info=e)
```

This outputs the error and traceback to both the screen and the logfile.

magnet_coils.py
```diff
@@ -0,0 +1,373 @@
...
+            ['circle' (str), radius (float, cm)]
+            For a rectangular cross-section, the list format is
+            ['rectangle' (str), width (float, cm), thickness (float, cm)]
+        logger (object): logger object.
+
+    Returns:
+        vol_ids (list of int): indices for magnet volumes.
+    """
+    # Extract cross-section parameters
+    shape_str = extract_cs(cross_section, logger)
```
gonuke, line 170:
```suggestion
    try:
        shape_str = extract_cs(cross_section, logger)
    except ValueError:
        logger.error(...)
```

parametric_stellarator.py
```diff
@@ -380,45 +512,80 @@ def parametric_stellarator(
...
-
-    # Check that h5m_export has an appropriate value
-    if h5m_export != None and h5m_export != 'Cubit' and h5m_export != 'Gmsh':
-        raise ValueError(
-            'h5m_export must be None or have a string value of \'Cubit\' or '
-            '\'Gmsh\''
-        )
+    except AssertionError as e:
+        logger.error(e.args[0])
+        raise e
```
connoramoreno, line 577: Note here that I used a different implementation for exception handling. Because there is no script level outside this function aside from the calling script (this is the main function), I couldn't put a try statement on the outside of this functionality. I didn't want to throw this check inside a different function (such as `stellarator_torus`) because I figured it didn't make sense to check it multiple times (as `stellarator_torus` gets called multiple times)

gonuke: OK - but I think you need to initialize the logger first - see a few lines below...

gonuke (changes requested):
One last tweak to make sure the logger is initialized before it'sused.

PR #5, opened 2023-06-19
Add source mesh modeling

connoramoreno:
Adds ability to model a source mesh based on plasma equilibrium data. The mesh is structured in confinement space and transformed to a tetrahedral mesh in real space.

### DIFF
--- a/source_mesh.py
+++ b/source_mesh.py
@@ -316,14 +316,14 @@ def create_vertices(vmec, mbc, s_list, theta_list, phi_list):
         # Determine vertex at magnetic axis, converting to cm
         vertex = np.array(vmec.vmec2xyz(s_list[0], theta_list[0], phi)) * m2cm
         # Append vertex to list
-        verts += [list(vertex)]
+        verts += [vertex]
         # Store s for vertex
         verts_s += [s_list[0]]
         for s in s_list[1:]:
             for theta in theta_list:
                 # Detemine vertices beyond magnetic axis in same toroidal angle
                 vertex = np.array(vmec.vmec2xyz(s, theta, phi)) * m2cm
-                verts += [list(vertex)]
+                verts += [vertex]
                 # Store s for vertex
                 verts_s += [s]
 
--- a/source_mesh.py
+++ b/source_mesh.py
@@ -27,7 +27,7 @@ def rxn_rate(s):
     return rr
 
 
-def source_strength(verts, verts_s, id0, id1, id2, id3):
+def source_strength(verts, verts_s, ids):
     """Computes neutron source strength for a tetrahedron using five-node
     Gaussian quadrature.
 
@@ -36,34 +36,29 @@ def source_strength(verts, verts_s, id0, id1, id2, id3):
             vertex in form [x (cm), y (cm), z (cm)].
         verts_s (list of float): list of closed flux surface indices for each
             vertex.
-        id0 (int): first tetrahedron index in MOAB canonical numbering system.
-        id1 (int): second tetrahedron index in MOAB canonical numbering system.
-        id2 (int): third tetrahedron index in MOAB canonical numbering system.
-        id3 (int): fourth tetrahedron index in MOAB canonical numbering system.
+        ids (list of int): tetrahedron vertex indices.
 
     Returns:
         ss (float): integrated source strength for tetrahedron.
     """
-    # Define vertices for tetrahedron
-    verts0 = verts[id0]
-    verts1 = verts[id1]
-    verts2 = verts[id2]
-    verts3 = verts[id3]
+    # Initialize list of coordinates for each tetrahedron vertex
+    tet_verts = []
+    # Initialize list of source strengths for each tetrahedron
+    ss_verts = []
 
-    # Compute fusion source density at each vertex
-    ss0 = rxn_rate(verts_s[id0])
-    ss1 = rxn_rate(verts_s[id1])
-    ss2 = rxn_rate(verts_s[id2])
-    ss3 = rxn_rate(verts_s[id3])
+    # Define vertices for tetrahedron
+    for id in ids:
+        tet_verts += [verts[id]]
+        ss_verts += [rxn_rate(verts_s[id])]
 
     # Define barycentric coordinates for integration points
-    bary_coords = [
+    bary_coords = np.array([
         [0.25, 0.25, 0.25, 0.25],
         [0.5, 1/6, 1/6, 1/6],
         [1/6, 0.5, 1/6, 1/6],
         [1/6, 1/6, 0.5, 1/6],
         [1/6, 1/6, 1/6, 0.5]
-    ]
+    ])
 
     # Define weights for integration points
     int_w = [-0.8, 0.45, 0.45, 0.45, 0.45, 0.45]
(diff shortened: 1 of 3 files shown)


source_mesh.py
```diff
@@ -0,0 +1,381 @@
...
+        id3 (int): fourth tetrahedron index in MOAB canonical numbering system.
+
+    Returns:
+        ss (float): integrated source strength for tetrahedron.
+    """
+    # Define vertices for tetrahedron
+    verts0 = verts[id0]
+    verts1 = verts[id1]
+    verts2 = verts[id2]
+    verts3 = verts[id3]
```
gonuke, line 51: Why not a numpy vector?

source_mesh.py
```diff
@@ -0,0 +1,381 @@
...
+
+    # Compute fusion source density at each vertex
+    ss0 = rxn_rate(verts_s[id0])
+    ss1 = rxn_rate(verts_s[id1])
+    ss2 = rxn_rate(verts_s[id2])
+    ss3 = rxn_rate(verts_s[id3])
```
gonuke, line 57: numpy vector?

source_mesh.py
```diff
@@ -0,0 +1,381 @@
...
+        [1/6, 1/6, 0.5, 1/6],
+        [1/6, 1/6, 1/6, 0.5]
+    ]
+
+    # Define weights for integration points
+    int_w = [-0.8, 0.45, 0.45, 0.45, 0.45, 0.45]
+    
+    # Interpolate source strength at integration points
+    ss_int_pts = []
+    for pt in bary_coords:
```
gonuke, line 73: numpy vector

source_mesh.py
```diff
@@ -0,0 +1,381 @@
...
+            verts2[2] - verts3[2]
+        ]
+    ]
+    
+    # Compute Jacobian of graph
+    Jac = np.linalg.det(T)
+    # Compute volume of tetrahedron
+    vol = np.abs(Jac)/6
+    # Compute source strength of tetrahedron
+    ss = vol*sum(i*j for i, j in zip(int_w, ss_int_pts))
```
gonuke, line 101: PEP8
```suggestion
    ss = vol * sum(i * j for i, j in zip(int_w, ss_int_pts))
```

source_mesh.py
```diff
@@ -0,0 +1,381 @@
...
+    """
+    # Initialize list of vertices in mesh
+    verts = []
+    # Initialize list of closed flux surface indices for each vertex
+    verts_s = []
+
+    # Compute vertices in Cartesian space
+    for phi in phi_list:
+        # Determine vertex at magnetic axis, converting to cm
+        vertex = np.array(vmec.vmec2xyz(s_list[0], theta_list[0], phi))*100
```
gonuke, line 278: define a `m2cm` constant?  also PEP8

source_mesh.py
```diff
@@ -0,0 +1,381 @@
...
+            for k, theta in enumerate(theta_list[:-1], 1):
+                # Define indices for hexahedron beyond center of plasma
+                tet_id0 = ma_id + s_offset + k
+                tet_id1 = ma_id + next_s_offset + k
+                tet_id2 = ma_id + next_s_offset + k + 1
+                tet_id3 = ma_id + s_offset + k + 1
+                tet_id4 = next_ma_id + s_offset + k
+                tet_id5 = next_ma_id + next_s_offset + k
+                tet_id6 = next_ma_id + next_s_offset + k + 1
+                tet_id7 = next_ma_id + s_offset + k + 1
```
gonuke, line 223: play with `numpy.ravel_multi_index()` to make this more readable by future you (or future me)

source_mesh.py
```diff
@@ -0,0 +1,381 @@
...
+        for j, s in enumerate(s_list[1:-1]):
+            # Define index at current closed magnetic flux surface
+            s_offset = j*num_theta
+            # Define index at next closed magnetic flux surface
+            next_s_offset = s_offset + num_theta
+
+            # Create tetrahedra for current hexahedron
+            for k, theta in enumerate(theta_list[:-1], 1):
+                # Define indices for hexahedron beyond center of plasma
+                tet_id0 = ma_id + s_offset + k
```
gonuke, line 216: Array instead of numbered variables

source_mesh.py
```diff
@@ -0,0 +1,381 @@
...
+                    tet_id2, tet_id1, tet_id3, tet_id6, strengths
+                )
+                create_tet(
+                    mbc, tag_handle, tet_set, mbc_verts, verts, verts_s,
+                    tet_id5, tet_id6, tet_id4, tet_id1, strengths
+                )
+                create_tet(
+                    mbc, tag_handle, tet_set, mbc_verts, verts, verts_s,
+                    tet_id3, tet_id1, tet_id4, tet_id6, strengths
+                )
```
gonuke, line 245: create a list and/or numpy array of these canonical orderings and loop over them

gonuke: Do this in a `create_tets_from_hex()` method

source_mesh.py
```diff
@@ -0,0 +1,381 @@
...
+                wedge_id2, wedge_id4, wedge_id0, strengths
+            )
+            create_tet(
+                mbc, tag_handle, tet_set, mbc_verts, verts, verts_s, wedge_id5,
+                wedge_id4, wedge_id2, wedge_id3, strengths
+            )
+            create_tet(
+                mbc, tag_handle, tet_set, mbc_verts, verts, verts_s, wedge_id0,
+                wedge_id2, wedge_id4, wedge_id3, strengths
+            )
```
gonuke, line 204: `create_tets_from_wedge()`?

gonuke (changes requested):
Thanks @connoramoreno - hopefully this gets more compact and readable with these changes

source_mesh.py
```diff
@@ -0,0 +1,420 @@
...
+            ]
+            # Initialize list of vertex indices for wedges
+            ids_wedge = []
+            # Define vertex indices in row-major order for wedges
+            for vertex in wedge_id_data:
+                ids_wedge += [
+                    int(np.ravel_multi_index(
+                        vertex[:3], [num_phi, num_s - 1, num_theta]
+                    )) + vertex[3]
+                ]
```
gonuke, line 243: This seems more complicated than it needs to be... let's discuss in a software meeting. Maybe it's because the array dimensions are not consistent because of the magnetic axis?

gonuke (changes requested):
Thanks for exploring the `ravel_index` version.  I can imagine that you find it even less readable :) and I might not disagree.  I think we need to talk it over in a s/w meeting.  There are some standard patterns for this kind of thing that I think I can share better in person.

source_mesh.py
connoramoreno: Yes, that's correct. I could change the structure so that there's a magnetic axis value for each theta angle (returns the same point) but that seems wasteful from a storage perspective

connoramoreno: Doing so would result in roughly `num_phi*num_theta` additional points

gonuke (approved):
A couple of little comments, but I'm going to merge anyway.  We can create issues to consider how we might address this going forward.

PR #6, opened 2023-07-14
Update README.md

gonuke:
Updated Readme with link to dependencies to make them easier to find.


README.md
connoramoreno, line 1: It would probably be a good idea to add CadQuery, MOAB (or PyMOAB), and Cubit to the list of dependencies. I can add those myself!

connoramoreno (commented):
Good idea to update the README! It's been a little lacking.

PR #9, opened 2023-07-25
Add toroidal and poloidal variation of component thickness

connoramoreno:
Adds ability for user to specify toroidal and poloidal variation of component thicknesses. Users specify component thicknesses at points in the toroidal and poloidal domain, and thickness values are linearly interpolated between these points to provide continuous variation.


parametric_stellarator.py
```diff
@@ -373,43 +375,106 @@ def offset_surface(vmec, theta, phi, offset, period_ext):
...
+    elif offset < 0:
+        raise ValueError(
+            'Offset must be greater than or equal to 0. Check thickness inputs '
+            'for negative values'
+        )
 
     return r
 
 
+def expand_offset(offset_vals, ang_list, ang_list_exp, num_ang_list):
```
gonuke, line 389: Perhaps we should use the existing SciPy capability for this: https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.RegularGridInterpolator.html#scipy-interpolate-regulargridinterpolator

gonuke (changes requested):
I made a couple of initial requests, but I think it will be good to discuss data structure choices for the now quite complex radial build data in a S/W meeting

parastell.py
```diff
@@ -558,44 +489,33 @@ def expand_ang(ang_list, num_ang):
...
-    for i, ang in enumerate(ang_list[:-1]):
+    for ang, next_ang in zip(ang_list[:-1], ang_list[1:]):
         # Compute difference between current and next angles
-        ang_diff = ang_list[i + 1] - ang
+        ang_diff = next_ang - ang
         # Compute ratio of angular difference and total angular extent
         ratio = ang_diff/ang_ext
         
         # Compute number of angles to interpolate
         n_ang = round(ratio*num_ang)
```
gonuke, line 507: There may be a clearer way to do this by calculating a `max_ang_diff = ang_ext/num_ang` outside of the loop, and then calculating `n_ang = ceil( (next_ang - ang) / max_ang_diff )`

gonuke (changes requested):
A small comment on the readability and robustness of your angle expansion.

Aside: I think we are at a tipping point for wanting to make this more object oriented!

parastell.py
```diff
@@ -465,13 +478,52 @@ def stellarator_torus(
...
+    # Compute total angular extent of supplied list
+    ang_ext = ang_list[-1] - ang_list[0]
+
+    # Compute average distance between angles to include in stellarator build
+    ang_diff_avg = ang_ext/num_ang
+    
+    # Loop over supplied angles
+    for ang, next_ang in zip(ang_list[:-1], ang_list[1:]):
+        # Compute number of angles to interpolate
+        n_ang = int(np.ceil((next_ang - ang)/ang_diff_avg))
```
gonuke, line 505: `ceil()` should always return an integer already

gonuke (changes requested):
One little change recommended here

gonuke:
I think this needs a rebase from #11 ?

parastell.py
connoramoreno: `np.ceil()` returns a float. I could use the math package instead to return an integer. I just didn't want to import a package for a single function, but I could just import `math.ceil()`.

connoramoreno:
Rebased on #11

parastell.py
gonuke: No, what you have is fine... odd choice for `numpy` to return a float version of the integer that this method evaluates

gonuke (approved):
Looks good

PR #10, opened 2023-07-27
Mesh integration

eitan-weinstein:
This PR integrates tetrahedral meshing into the magnet coil creation, so that it can be directly incorporated into the parametric stellarator, if desired. `ExampleScript.py` was also modified to allow for the meshing.


parametric_stellarator.py
```diff
@@ -206,6 +208,16 @@ def exports(export, components, magnets, logger):
...
     
+    if magnets['meshing']:
+        # Get current working directory
+        cwd = os.getcwd()      
+        # Exodus export
+        exo_path = f'{cwd}/coil_mesh.exo'
+        cubit.cmd(f'export mesh "{exo_path}"')
+        # h5m conversion
+        h5m_path = f'{cwd}/coil_mesh.h5m'
+        h5m_conversion = subprocess.run(f'mbconvert {exo_path} {h5m_path}', shell = True)
```
gonuke, line 219: I think we can do this with pymoab commands by just reading and writing a file in different formats, possibly in a new MOAB instance but maybe just into a new meshset?  It will be lighter weight than launching a subprocess.

coils_tet_mesh.py
```diff
@@ -0,0 +1,189 @@
...
+    # GEOMETRY AND MESH CREATION
+
+    # Start Cubit - this step is key
+    cubit.init(['cubit', '-nojournal'])
+
+    # Import coils
+    cubit.cmd(f'import step "{coils_path}" heal')
+    cubit.cmd('volume  size auto factor 5')
+
+    # For-loop iterates over all 48 coils included in the coils.step file
```
gonuke, line 42: Comment still refers to 48 coils

coils_tet_mesh.py
```diff
@@ -0,0 +1,189 @@
...
+    for coil_num in np.arange(1,total_coil_count):
+        
+        # Create the mesh
+        cubit.cmd(f'volume {coil_num} scheme tetmesh')
+        cubit.cmd(f'volume {coil_num} scheme tetmesh')
```
gonuke, line 47: Repeated line... does it do anything?

coils_tet_mesh.py
```diff
@@ -0,0 +1,189 @@
...
+        cubit.cmd(f'mesh volume {coil_num}')
+
+        # Set log file path and initiate logging
+        log_split = log_general_path.split('.')
```
gonuke, line 51: Do this once before the loop, perhaps name the components for clarity something like `log_prefix` and `log_ext`?

coils_tet_mesh.py
```diff
@@ -0,0 +1,189 @@
...
+        log_path = f"{log_split[0]}_{coil_num}.{log_split[1]})"
+        cubit.cmd(f'logging on file "{log_path}"')
+
+        # Mesh quality metrics
+        quality_metric_list = ['shape', 'aspect ratio', 'condition no.', 'distortion', 'element volume',
+                                'equiangle skew', 'equivolume skew','inradius', 'jacobian', 
+                                'normalized inradius', 'relative size', 'scaled jacobian', 'shape and size']
```
gonuke, line 58: Define this once before the loop

coils_tet_mesh.py
```diff
@@ -0,0 +1,189 @@
...
+            # Break the loop if the ending cue is not found
+            if end_index == -1:
+                break
+
+            # Extract the range between the starting and ending index
+            range_text = text[start_index : end_index + 1]
+            ranges.append(range_text)
+
+            # Update the starting index for the next iteration
+            start_index = end_index + 1
```
gonuke, line 98: Maybe this can be a function to reduce cognitive load

gonuke: Arguably, the file reading should perhaps be outside the function for separation of concerns

coils_tet_mesh.py
```diff
@@ -0,0 +1,189 @@
...
+            row_extraction = row_extraction[:5]
+            row_extraction[0][0] = row_extraction[0][0].split('-\n')[1]
+
+            data_lists = [function_names, averages, std_devs, minima, maxima]
+            indices = np.arange(5)
+            for index, data_list in zip(indices, data_lists):
+                if index == 0:
+                    data_list.append(row_extraction[index][0])
+                else:
+                    data_list.append(float(row_extraction[index][0]))
```
gonuke, line 128: Maybe another function to reduce cognitive load

gonuke: We'll also want to review whether there are better ways to do this parsing....

coils_tet_mesh.py
```diff
@@ -0,0 +1,189 @@
...
+
+        # Recreate the geometry and mesh with the best_analytic
+        cubit.cmd('reset')
+        cubit.cmd(f'import step "{coils_path}" heal')
+        cubit.cmd('volume  size auto factor 5')
+        cubit.cmd(f'volume {coil_num} scheme tetmesh')
+        cubit.cmd(f'volume {coil_num} scheme tetmesh')
+        cubit.cmd(f'mesh volume {coil_num}')
+        cubit.cmd(f'mesh volume {coil_num}')
+        cubit.cmd(f'quality volume {coil_num} {best_analytic} global draw mesh list')
```
gonuke, line 179: I'm not sure this is doing anything meaningful...

gonuke (changes requested):
A few initial thoughts... the integration with the @connoramoreno  code is good. For the mesh analysis code... I've suggested some things to break up the code.

eitan-weinstein:
Thanks for the comments, Paul. I'll get on them now and see if I can implement these improvements.

parametric_stellarator.py
eitan-weinstein: I looked in the MOAB bitbucket for a pymoab version of mbconvert, but could not find anything. Is there a specific function that you have in mind to replace the subprocess command?

gonuke: Something like:
```
myFileset = pymoab.EntitySet()
myMoab.read_file('file.exo', fileset = myFileset)
myMoab.write_file('file.h5m', fileset = myFileset)
```

eitan-weinstein:
Sorry about the whirlwind of commits in the past few minutes. I had made an accidental commit when trying to update the `coils_tet_mesh.py` script and then made a few compounding mistakes on GitHub trying to fix it all. It looks like it should be updated properly now. My newest change incorporates the PyMOAB commands in the mesh file conversion, as well as the export of a .csv file containing the mesh quality analytics for each coil.

gonuke:
somewhere you have lost the integration of meshing within the main package

eitan-weinstein:
Alright, the integration has been re-incorporated and updated to include the PyMOAB commands.

parametric_stellarator.py
```diff
@@ -205,6 +207,27 @@ def exports(export, components, magnets, logger):
         logger.info('Exporting STEP files...')
         for name, comp in components.items():
             cq.exporters.export(comp['solid'], name + '.step')
+            
+        # Conditionally export tetrahedral meshing
+        if magnets['meshing']:
+            # Get current working directory
+            cwd = os.getcwd()      
+            # Exodus export
+            exo_path = f'{cwd}/coil_mesh.exo'
```
gonuke, line 216: Define the `base_name` of the file
```
base_name = 'coil_mesh'
exo_path = f'{file_path}/{base_name}.exo'
```
and use that in both places.  This will facilitate changing it in the future.

parametric_stellarator.py
```diff
@@ -205,6 +207,27 @@ def exports(export, components, magnets, logger):
         logger.info('Exporting STEP files...')
         for name, comp in components.items():
             cq.exporters.export(comp['solid'], name + '.step')
+            
+        # Conditionally export tetrahedral meshing
+        if magnets['meshing']:
+            # Get current working directory
+            cwd = os.getcwd()
```
gonuke, line 214:
```suggestion
            file_path = os.getcwd()      
```

parametric_stellarator.py
```diff
@@ -205,6 +207,27 @@ def exports(export, components, magnets, logger):
...
+            exodus_set = mb.create_meshset()
+            mb.load_file(exo_path, exodus_set)
+            # Create a new meshset for each coil iteration
+            coil_meshset = mb.create_meshset()
+            # Add the current coil's mesh to the meshset
+            mb.add_entities(coil_meshset, [exodus_set])
+            # Set .h5m path
+            h5m_path = f'{cwd}/coil_mesh.h5m'
+            # Write the current coil's meshset to an individual .h5m file
+            mb.write_file(h5m_path, [coil_meshset])
```
gonuke, line 230:
```suggestion
            # Set .h5m path
            h5m_path = f'{file_path}/{base_name}.h5m'
            # Write the current coil's meshset to an individual .h5m file
            mb.write_file(h5m_path, [exodus_set])
```

coils_tet_mesh.py
gonuke, line 1: Add a test for `__main__` (https://www.tutorialspoint.com/What-does-the-if-name-main-do-in-Python#:~:text=A%20Python%20programme%20uses%20the,is%20imported%20as%20a%20module.) that runs the `coils_tet_mesh` method when this file is run from the command-line.

BONUS: add `argparse` to get command line arguments (https://docs.python.org/3/library/argparse.html)

coils_tet_mesh.py
```diff
@@ -0,0 +1,277 @@
...
+    cubit.init(['cubit', '-nojournal'])
+
+    # Import coils
+    cubit.cmd(f'import step "{coils_path}" heal')
+    cubit.cmd('volume  size auto factor 5')
+
+    # Split log path
+    log_split = log_general_path.split('.')
+    log_prefix = log_split[0]
+    log_ext = log_split[1]
```
gonuke, line 175: might be easier with the `pathlib` package/module??

coils_tet_mesh.py
```diff
@@ -0,0 +1,277 @@
...
+        # Remove rows where the standard deviation equals 0
+        # Found this for a few cases and it appeared too implausible to keep
+        quality_metrics_df = quality_metrics_df.loc[quality_metrics_df['Standard Deviation'] != 0]
+
+        # Sort rows by highest average quality metric and select the highest
+        quality_metrics_df = quality_metrics_df.sort_values('Average', ascending=False)
+        best_analytic = quality_metrics_df['Function Name'][0]
+
+        # Override the current mesh analytic with the best quality metric
+        cubit.cmd(f'quality volume {coil_num} {best_analytic} global draw mesh list')
```
gonuke, line 233: May not be necessary?? We have this in the log file and/or data frame already

coils_tet_mesh.py
```diff
@@ -0,0 +1,277 @@
...
+        cubit.cmd(f'mesh volume {coil_num}')
+
+        # Set log file path and initiate logging
+        log_path = f"{log_prefix}_{coil_num}.{log_ext}"
+        cubit.cmd(f'logging on file "{log_path}"')
+
+        # Run mesh quality metrics for each relevant metric
+        for metric in quality_metric_list:
+            qual_command = f'quality volume {coil_num} {metric} global draw mesh list'
+            mesh_analytics = cubit.cmd(qual_command)
```
gonuke, line 202: Separation of concerns & cognitive load:
Make this a function that creates and does the quality measures on each mesh

coils_tet_mesh.py
```diff
@@ -0,0 +1,277 @@
...
+        # Compile the extracted tabular data into a Pandas data frame
+        quality_metrics_df = pd.DataFrame({'Function Name': function_names,
+                                           'Average': averages,
+                                           'Standard Deviation': std_devs,
+                                           'Minimum': minima,
+                                           'Maximum': maxima})
+
+        # Remove rows where the standard deviation equals 0
+        # Found this for a few cases and it appeared too implausible to keep
+        quality_metrics_df = quality_metrics_df.loc[quality_metrics_df['Standard Deviation'] != 0]
```
gonuke, line 226: Separation of concerns & cognitive load:
Make this a method that builds & cleans the dataframe

coils_tet_mesh.py
```diff
@@ -0,0 +1,277 @@
...
+    mb.load_file(exo_path, exodus_set)
+    
+    # Create a new meshset for each coil iteration
+    coil_meshset = mb.create_meshset()
+
+    # Add the current coil's mesh to the meshset
+    mb.add_entities(coil_meshset, [exodus_set])
+
+    # Write the current coil's meshset to an individual .h5m file
+    mb.write_file(h5m_path, [coil_meshset])
```
gonuke, line 261: Method to export the mesh - make it optional based on the user input

gonuke: Also update as in the parametric stellarator code

coils_tet_mesh.py
```diff
@@ -0,0 +1,277 @@
...
+    top_results.rename(columns = {'index' : 'Coil Number'}, inplace = True)
+
+    # Fix indexing for coil numbers
+    top_results = top_results.transpose()
+    top_results.iloc[0] = np.arange(1, total_coil_count + 1)
+    top_results = top_results.transpose()
+    print(top_results.head())
+
+    # Export the top_results data frame as a .csv
+    top_results.to_csv(csv_path)
```
gonuke, line 277: Function to write CSV - optional?
Separation of concerns: separate analysis from I/O

coils_tet_mesh.py
gonuke: Consider a method that can be reused here and in `parametric_stellarator` for mesh export as H5M.

gonuke (changes requested):
Looks great! The integrated meshing is great, still some more refining on the stand-alone analysis.

parametric_stellarator.py
```diff
@@ -205,6 +207,25 @@ def exports(export, components, magnets, logger):
...
+            cubit.cmd(f'export mesh "{exo_path}"')
+            # Initialize the MOAB core instance
+            mb = core.Core()
+            # Load the EXODUS file
+            exodus_set = mb.create_meshset()
+            mb.load_file(exo_path, exodus_set)
+            # Set .h5m path
+            h5m_path = f'{file_path}/{base_name}.h5m'
+            # Write the current coil's meshset to an individual .h5m file
+            mb.write_file(h5m_path, [exodus_set])
```
gonuke, line 228: A couple of fine tuning thoughts related to best practice, readability and cognitive load:
- Consider whether all the comments are necessary; good choice of variable names and code structure can make comments redundant and interrupt the flow reading a snippet of code. Comments can often be reserved for explain it more complicated algorithmic choices.
- If the two filename definitions are adjacent, it can be more readable by their association to each other as well as removing them from the "action" flow of the read/write steps

gonuke (changes requested):
Some quick feedback - very much in the fine tuning style aspects

gonuke (changes requested):
A few more ways to streamline and simplify some of the algorithms.

I think we perhaps should have sat down to discuss the cleanest way to extract data from the logs - it seems far more complicated than it probably needs to be.

There may be some value in separating the `coils_tet_mesh.py` script from the rest of this PR.

PR #18, opened 2023-11-03
Add install guide

### LINKED ISSUE #17: installation instructions
Hi all,

I'm interested in trying to run parastell and understand its capabilities better. Do you have any installation instructions?

Thanks!

Edgar-21:
Fixes #17

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,44 +11,21 @@ This tool depends on:
 - [cad-to-dagmc](https://pypi.org/project/cad-to-dagmc/)
 - [Gmsh](https://pypi.org/project/gmsh/)
 
-## Install with Cubit 2023.11
+## Install with Mamba
 
-### Parastell
-Download and extract repository for parastell using
+This guide will use the mamba package manager to install dependencies in a conda environment. Conda allows for easily installing and switching between different versions of software packages through the use of [environments](https://conda.io/projects/conda/en/latest/user-guide/concepts/environments.html).
 
-```
-git clone git@github.com:svalinn/parastell.git
-```
+If you have not already installed conda, you can use one of the following installers to do so: [Miniconda](https://docs.conda.io/en/latest/miniconda.html), [Anaconda](https://www.anaconda.com/), or [Miniforge](https://github.com/conda-forge/miniforge).
 
-or download the zip from the repository home page. Once extracted, add the directory to the `PYTHONPATH` by adding a line to the .bashrc file like the following:
+Mamba is available through conda with `conda install -c conda-forge mamba`. Begin by creating a new conda environment with mamba installed.
 
 ```
-export PYTHONPATH=$PYTHONPATH:$HOME/parastell/
+conda create -n parastell_env
+conda activate parastell_env
+conda install -c conda-forge mamba
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,10 +1,77 @@
...
 - [CadQuery](https://cadquery.readthedocs.io/en/latest/installation.html)
 - [MOAB](https://bitbucket.org/fathomteam/moab/src/master/)
 - [Coreform Cubit](https://coreform.com/products/downloads/) or [Cubit](https://cubit.sandia.gov/downloads/) by Sandia National Laboratories
+- [cad-to-dagmc](https://pypi.org/project/cad-to-dagmc/)
+- [Gmsh](https://pypi.org/project/gmsh/)
+
+## Install with Cubit 2023.11
+
+### Parastell
+Download and extract repository for parastell using
```
gonuke, line 17: Normally, I would list this step last... install all the dependencies first.

README.md
```diff
@@ -1,10 +1,77 @@
...
+Install ParaStell dependencies available on conda-forge.
+
+```
+mamba install -c conda-forge cadquery moab
+```
+
+Install pystell_uw dependencies available on conda-forge.
+
+```
+mamba install -c conda-forge matplotlib
```
gonuke, line 63: Is there a reason to use the `condo-forge` channel for this?

README.md
```diff
@@ -1,10 +1,77 @@
...
+
+If you do not have a Cubit license, you may be able to get one through [Cubit Learn](https://coreform.com/products/coreform-cubit/free-meshing-software/) at no cost.
+
+### Create a Conda Environment for ParaStell
+This guide uses the mamba package manager. If the mamba package manager is not available, using conda in place of mamba in the following commands should also work.
+
+```
+mamba create -n parastell_env
+mamba activate parastell_env
+``````
```
gonuke, line 51: This is not strictly necessary, so it may be useful to describe why it is recommended.

README.md
```diff
@@ -1,10 +1,77 @@
...
+mamba install -c conda-forge cadquery moab
+```
+
+Install pystell_uw dependencies available on conda-forge.
+
+```
+mamba install -c conda-forge matplotlib
+```
+
+Pip install remaining pystell_uw dependency.
```
gonuke, line 66: Are these pip installs also confined to this condo env?  I don't actually know.  If they are not, then probably best to do them before the condo env and installs.

README.md
```diff
@@ -1,10 +1,77 @@
...
+
+```
+pip install netCDF4
```
gonuke, line 69: Why do we need this? Isn't handled as a dependency of something else? (E.g. Moab or gmsh?)

gonuke (changes requested):
Thanks for putting this together @Edgar-21.

I think the most logical order is:
* pip installs (move to 3rd spot if pip installs are confined to a particular condo env)
* condo env
* condo installs
* Cubit
* vmec 
* parastell

README.md
Edgar-21: describe the benefit of using a conda environment?

README.md
Edgar-21: i believe if pip is ran inside the conda env then those packages will only be available in that environment, which I think is what we want, I will make it clearer that this takes place in the conda env

README.md
Edgar-21: from testing it is not a dependency of anything else and needs to be specified

gonuke (approved):
Thanks @Edgar-21 - this should do the trick

PR #19, opened 2023-11-10
Add support for modeling a toroidal segment and change offset algorithm

connoramoreno:
This pull request changes the offset algorithm from a 3-D offset along the plasma surface normal and a 2-D offset along the poloidal profile normal. Additionally, support is added for modeling geometries less than 360 degrees. The support includes modeling a portion of the source mesh and magnets.

### DIFF
--- a/magnet_coils.py
+++ b/magnet_coils.py
@@ -290,12 +290,11 @@ def clean_mag_data(filaments, tor_ext, r_avg, mag_len):
         com = np.average(fil, axis = 0)
         # Compute toroidal angle of each point in filament
         phi_pts = np.arctan2(fil[:,1], fil[:,0])
+        # Ensure angles are positive
+        phi_pts = (phi_pts + 2*np.pi) % (2*np.pi)
         # Compute bounds of toroidal extent of filament
         min_phi = np.min(phi_pts)
         max_phi = np.max(phi_pts)
-        # Ensure angles are positive
-        min_phi = (min_phi + 2*np.pi) % (2*np.pi)
-        max_phi = (max_phi + 2*np.pi) % (2*np.pi)
 
         # Determine if filament toroidal extent overlaps with that of model
         if (
@@ -310,6 +309,7 @@ def clean_mag_data(filaments, tor_ext, r_avg, mag_len):
 
     # Compute toroidal angles of filament centers of mass
     phi_arr = np.arctan2(com_list[:,1], com_list[:,0])
+    phi_arr = (phi_arr + 2*np.pi) % (2*np.pi)
 
     # Sort filaments by toroidal angle
     sorted_fils = [x for _,x in sorted(zip(phi_arr, reduced_fils))]
--- a/magnet_coils.py
+++ b/magnet_coils.py
@@ -3,6 +3,55 @@ import cubit
 import numpy as np
 
 
+def mesh_magnets(vol_ids):
+    """Creates tetrahedral mesh of magnet volumes.
+
+    Arguments:
+        vol_ids (list of int): indices for magnet volumes.
+    """
+    # Loop over magnet indices
+    for vol in vol_ids:
+        # Create scheme and meshes
+        cubit.cmd(f'volume {vol} scheme tetmesh')
+        cubit.cmd(f'mesh volume {vol}')
+
+
+def cut_mags(tor_ext, vol_ids):
+    """Cuts magnet volumes such that only the specified toroidal extent is
+    included.
+    
+    Arguments:
+        tor_ext (float): toroidal extent to model (deg).
+        vol_ids (list of int): indices for magnet volumes.
+    
+    Returns:
+        vol_ids (list of int): updated indices for magnet volumes.
+    """
+    # Create surface
+    cubit.cmd('create surface rectangle width 4000 yplane')
+    # Extract surface index
+    surf_id = cubit.get_last_id("surface")
+    # Move surface
+    cubit.cmd(f'move Surface {surf_id} x 2000')
+    # Sweep surface
+    cubit.cmd(f'sweep surface {surf_id} zaxis angle {tor_ext}')
+    # Extract volume index
+    sweep_id = cubit.get_last_id("volume")
+
+    # Intersect magnet volumes with sweep volume
+    cubit.cmd(
+        'intersect volume ' + ' '.join(str(i) for i in vol_ids)
+        + f' {sweep_id}'
+    )
+
+    # Extract index of most recently created volume
+    last_id = cubit.get_last_id("volume")
+    # Compute magnet volume indices
+    vol_ids = range(sweep_id + 1, last_id + 1)
+
+    return vol_ids
+
+
 def unit_vector(vec):
     """Normalizes given vector.
 
@@ -147,7 +196,7 @@ def create_magnets(filaments, cross_section, meshing, logger):
     """Creates magnet coil solids.
     
     Arguments:
-        filaments (list of list of list of float): list filament coordinates.
+        filaments (list of list of list of float): list of filament coordinates.
             Each filament is a list of coordinates.
         cross_section (list or tuple of str, float, float): coil cross-section
             definition. Note that the cross-section shape must be either a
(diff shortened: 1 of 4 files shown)


parastell.py
```diff
@@ -699,11 +645,22 @@ def parastell(
     wall_s = build['wall_s']
     # Extract radial build
     radial_build = build['radial_build']
-
     # Extract array dimensions
     n_phi = len(phi_list)
     n_theta = len(theta_list)
 
+    # Compute toroidal extent of build in degrees
+    tor_ext = phi_list[-1] - phi_list[0]
```
gonuke, line 653: per previous comment, I think it makes sense to convert this to radians ASAP and everything else in radians internally

parastell.py
```diff
@@ -780,23 +735,32 @@ def parastell(
...
         cubit.init([
             'cubit',
             '-nojournal',
             '-nographics',
             '-information', 'off',
-            '-warning', 'off'
+            '-warning', 'off',
+            '-commandplugindir',
+            cubit_dir
         ])
```
gonuke, line 757: this should probably happen as early as possible so that you can fail early if cubit is missing or mis-configured

source_mesh.py
```diff
@@ -192,14 +188,16 @@ def get_vertex_id(vert_ids, num_phi, num_s, num_theta):
...
     
     Returns:
         id (int): vertex index in row-major order as stored by MOAB
     """
     # Compute index offset from magnetic axis
-    if vert_ids[0] == num_phi - 1:
-        # Ensure logical connectivity at final toroidal angle
-        ma_offset = 0
+    if tor_ext == 360.0:
+        if vert_ids[0] == num_phi - 1:
```
gonuke, line 198: Should these conditions be combined with an `and`?  right now, `ma_offset` is not determined if `tor_ext == 360` and `vert_ids[0] != num_phi -1`

source_mesh.py
```diff
@@ -324,31 +329,35 @@ def create_vertices(vmec, mbc, num_s, num_theta, num_phi):
...
+        num_verts = num_phi * (1 + ((num_s - 1) * (num_theta - 1)))
     # Initialize n x 3 array of vertex coordinates
     verts = np.zeros((num_verts, 3))
     # Initialize array of vertex closed flux surface indices
     verts_s = np.zeros(num_verts)
 
     # Compute regular spacing between toroidal angles in confinement space to
     # be included in mesh
-    delta_phi = 2*np.pi/(num_phi - 1)
+    if tor_ext == 360.0:
```
gonuke, line 343: I think it makes sense to deal with both impacts of `tor_ext == 360` in the same place, if possible.

gonuke (changes requested):
There's a lot going on here @connoramoreno - A lot of comments are about trying to streamline and simplify.

magnet_coils.py
```diff
@@ -3,6 +3,55 @@
...
+        vol_ids (list of int): indices for magnet volumes.
+    """
+    # Loop over magnet indices
+    for vol in vol_ids:
+        # Create scheme and meshes
+        cubit.cmd(f'volume {vol} scheme tetmesh')
+        cubit.cmd(f'mesh volume {vol}')
+
+
+def cut_mags(tor_ext, vol_ids):
```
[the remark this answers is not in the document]
connoramoreno: Yes, some magnets will only exist in part.

The method that filters the filaments that are inside the toroidal extent is there to reduce the number of sweep operations and the operations required for the intersection routine. My thought was it didn't make sense to generate all magnets just to (potentially) remove most of them.

parastell.py
```diff
@@ -292,80 +294,43 @@ def graveyard(vmec, offset, components, logger):
...
     """Stellarator offset surface root-finding problem.
 
     Arguments:
         vmec (object): plasma equilibrium object.
         s (float): normalized magnetic closed flux surface value.
-        zeta (float): toroidal angle being solved for (rad).
+        phi (float): toroidal angle being solved for (rad).
         theta (float): poloidal angle of interest (rad).
         offset (float): total offset of layer from plamsa (m).
+        plane_norm (array of float): normal direction of toroidal plane.
```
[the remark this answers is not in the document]
connoramoreno: My thought was to calculate it once per toroidal angle in the `stellarator_torus` method, since it will be the same for each `phi`, rather than compute it repeatedly for each point in `offset_point`

parastell.py
```diff
@@ -453,6 +403,14 @@ def stellarator_torus(
         # Convert toroidal (phi) angle from degrees to radians
         phi = np.deg2rad(phi)
 
+        # Compute radial direction of poloidal profile
+        x = np.cos(phi)
+        y = np.sin(phi)
+        dir = np.array([x, y, 0])
+        # Compute norm of poloidal profile
+        plane_norm = np.cross(dir, np.array([0, 0, 1]))
```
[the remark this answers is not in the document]
connoramoreno: Good point

parastell.py
```diff
@@ -516,13 +474,12 @@ def expand_ang(ang_list, num_ang):
...
     # Compute average distance between angles to include in stellarator build
-    ang_diff_avg = ang_ext/num_ang
+    ang_diff_avg = ang_ext/(num_ang - 1)
     
     # Loop over supplied angles
     for ang, next_ang in zip(ang_list[:-1], ang_list[1:]):
         # Compute number of angles to interpolate
-        n_ang = int(np.ceil((next_ang - ang)/ang_diff_avg))
-
+        n_ang = math.ceil((next_ang - ang)/ang_diff_avg)
```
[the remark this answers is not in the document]
connoramoreno: Either works as intended, `math.ceil` is slightly cleaner

connoramoreno:
I've been working on a new commit, should be able to push by EOD

source_mesh.py
```diff
@@ -61,23 +61,18 @@ def source_strength(verts, verts_s, ids):
...
     # Define weights for integration points
-    int_w = [-0.8, 0.45, 0.45, 0.45, 0.45, 0.45]
+    int_w = np.array([-0.8, 0.45, 0.45, 0.45, 0.45])
     
     # Interpolate source strength at integration points
-    ss_int_pts = []
-    for pt in bary_coords:
-        ss_int = np.dot(pt, ss_verts)
-        ss_int_pts.append(ss_int)
+    ss_int_pts = np.dot(bary_coords, ss_verts)
```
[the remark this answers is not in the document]
connoramoreno: `np.dot` can accept lists and convert them to arrays internally

magnet_coils.py
```diff
@@ -276,6 +256,161 @@ def create_magnets(filaments, cross_section, meshing, logger):
...
+        # Compute filament center of mass
+        com = np.average(fil, axis = 0)
+        # Compute toroidal angle of each point in filament
+        phi_pts = np.arctan2(fil[:,1], fil[:,0])
+        # Compute bounds of toroidal extent of filament
+        min_phi = np.min(phi_pts)
+        max_phi = np.max(phi_pts)
+        # Ensure angles are positive
+        min_phi = (min_phi + 2*np.pi) % (2*np.pi)
+        max_phi = (max_phi + 2*np.pi) % (2*np.pi)
```
gonuke, line 298: Don't you have to do this before taking the min/max?  it seems you could have something that is less than 0 get identified as the minimum and then become the maximum once you've done this operation.

gonuke (changes requested):
One question here about shifting angles to be in the positive space before vs after max/min

magnet_coils.py
connoramoreno: Because `min_phi` and `max_phi` are being tested independently against `min_rad` and `max_rad`, it ends up not making a difference functionally but ensuring angles are positive before extracting the minimum and maximum seems like the cleaner way to do it.

connoramoreno: New commit incoming, just testing

gonuke:
Please do a quick rebase and then I'm happy to merge @connoramoreno

gonuke:
BTW, testing would have been helpful to resolve this last question

gonuke:
Thanks @connoramoreno - this looks good now and we'll move forwards on some of the other PRs

gonuke (approved):

PR #21, opened 2024-01-24
Fix cubit export

### LINKED ISSUE #20: Version of Coreform Cubit for exporting DAGMC
What version of Coreform Cubit is intended to be used? Using 2023.11 on Linux, it does not seem to have 'export dagmc'. It has 'export cf_dagmc', but for some reason even when updating to use that, it errors out in the call in cubit_export

Edgar-21:
added plugin directory to cubit init, changed import order, set meshing: False in magnets in the example script since meshing the magnets can be a failure point. Renamed a variable to match the name that gets used (file_path -> cwd). Resolves #20

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,11 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+*.csv
+*.step
+*.h5m
+*.txt
+*.nc
+*.jou
+*.ods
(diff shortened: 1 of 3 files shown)


gonuke (approved):
Thanks @Edgar-21 - this looks good to me.  One day we'll need to figure out testing...

PR #29, opened 2024-02-01
Magnet volumetric mesh H5M output with Cubit

connoramoreno:
Changes the Cubit export of the magnet tetrahedral mesh directly to H5M format, rather than a Cubit export in Exodus format and subsequent conversion to H5M via MOAB. Note that direct export to H5M via Cubit requires Cubit version 2023.11.

### DIFF
--- a/ExampleScript.py
+++ b/ExampleScript.py
@@ -1,5 +1,4 @@
 import parastell
-import logging
 
 
 # Define plasma equilibrium VMEC file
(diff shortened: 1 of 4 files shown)


gonuke:
I think it's fine to have this based on #19.  You may want to put it in draft until we merge #19, and you'll want to rebase this branch as we make changes to the branch for #19 during review.

connoramoreno:
The rebase is getting overly complicated so I'm going to close this and open a new pull request based on the newest version of main since this is such a small adjustment.

PR #30, opened 2024-02-02
Add native cubit meshing option

Edgar-21:
an additional flag `"native_meshing:bool` has been added to the the export dict along with the `"anisotropic_ratio":Float` and the `"deviation_angle":Float` parameters for the native cubit meshing. Docstrings in parastell and cubit_export have been updated. Default export_def dict has been updated. cubit_export will now check the "native_meshing" flag and either use the new native workflow, or the legacy DAGMC workflow


ExampleScript.py
```diff
@@ -148,6 +151,6 @@
 # Create stellarator
 parastell.parastell(
     plas_eq, num_periods, build, gen_periods, num_phi, num_theta,
-    magnets = magnets, source = source,
```
gonuke, line 151: Perhaps you turned off the magnets and source for testing, but I think we should turn them back as we merge this PR.

parastell.py
```diff
@@ -93,36 +95,94 @@ def cubit_export(components, export, magnets):
...
+                cubit.cmd('set duplicate block elements off')
+                cubit.cmd("block " + str(block_number) + " add volume " + str(vol))
+            
+            # assign magnet material to block
+            cubit.cmd("block " + str(block_number) + " material " + ''.join(("\'",magnet_h5m_tag,"\'")))
+        
+        #mesh the model
+        cubit.cmd("set trimesher coarse on ratio " + str(anisotropic_ratio) + " angle " + str(deviation_angle))
+        cubit.cmd("surface all scheme trimesh")
+        cubit.cmd("mesh surface all")
```
gonuke, line 145: It may make sense to put all this in a function to enhance readability?

parastell.py
```diff
@@ -93,36 +95,94 @@ def cubit_export(components, export, magnets):
...
+
+        #export dagmc file
+        cubit.cmd(f'export cf_dagmc "{cwd + "/dagmc.h5m"}" overwrite')
+
+    else:
+
+        # Extract Cubit export parameters
+        facet_tol = export['facet_tol']
+        len_tol = export['len_tol']
+        norm_tol = export['norm_tol']
```
gonuke, line 155: Perhaps move these closer to where they are used... or maybe do without and just use `export['facet_tol']` (etc) in the two places each are used...

parastell.py
```diff
@@ -93,36 +95,94 @@ def cubit_export(components, export, magnets):
...
+            len_tol_str = f'length_tolerance {len_tol}'
+        if norm_tol is not None:
+            norm_tol_str = f'normal_tolerance {norm_tol}'
+        
+        # DAGMC export
         cubit.cmd(
-            f'group "mat:{magnet_h5m_tag}" add volume '
-            + " ".join(str(i) for i in magnets['vol_id'])
+            f'export dagmc "dagmc.h5m" {facet_tol_str} {len_tol_str} {norm_tol_str}'
+            f' make_watertight'
```
gonuke, line 184: It might make sense to put this in a function

gonuke (changes requested):
Thanks @Edgar-21 - just a few high level requests, recommendations

ExampleScript.py
Edgar-21: agreed

parastell.py
Edgar-21: moved each export method to a local function

parastell.py
Edgar-21: moved closer to where they are used, left the rest as is for now, I can change it to use the dict if so desired, however

parastell.py
Edgar-21: its now in the local function `legacy_export()`, though I am a bit unclear about which part specifically you meant to be in a function

parastell.py
```diff
@@ -76,10 +83,86 @@ def cubit_export(components, export, magnets):
             For a rectangular cross-section, the list format is
             ['rectangle' (str), width (float, cm), thickness (float, cm)]
     """
-    # Extract Cubit export parameters
-    facet_tol = export['facet_tol']
-    len_tol = export['len_tol']
-    norm_tol = export['norm_tol']
+    def legacy_export():
```
gonuke, line 86: I assume that because this function is defined inside the scope of the export function, that it doesn't need to pass all the variables....?

gonuke (commented):
Just a quick question

parastell.py
Edgar-21: yeah, local functions share the variables of the parent in a similar manner to how functions can access global variables, at least thats how I understand it

gonuke (approved):
Thanks @Edgar-21! This looks good and it will be good to get this out there for others.

PR #31, opened 2024-02-06
Add support for NWL calculations

### LINKED ISSUE #22: PR existing tool to read surface source data and generate NWL map
Create a pull-request with NWL tools ASAP with existing capability.

connoramoreno:
Adds a script to compute NWL and create all corresponding geometry, source, etc. For now, this functionality is included in a separate directory but this organization is free to change. Note that there are three primary functionalities in the main script `NWL.py`, reflected by the three new example scripts `NWL_geom.py`, `NWL_transport.py`, and `NWL_plot.py`. We should discuss this structure but it is primarily due to the (not strict) necessity of separate `conda` environments for ParaStell and OpenMC. Cubit (a dependency of ParaStell) is built with an HDF5 version that does not play well with `conda` environments containing OpenMC. Perhaps there is some solution we can figure out that would remedy this issue.

Fixes #22

### DIFF
--- a/NWL/NWL.py
+++ b/NWL/NWL.py
@@ -99,13 +99,14 @@ def NWL_geom(
     from scipy.interpolate import RegularGridInterpolator
     import os
     import inspect
+    from pathlib import Path
     
     export_dict = export_def.copy()
     export_dict.update(export)
     
     if export_dict['h5m_export'] == 'Cubit':
         cubit_dir = os.path.dirname(inspect.getfile(cubit))
-        cubit_dir = cubit_dir + '/plugins/'
+        cubit_dir = Path(cubit_dir) / Path('plugins')
         cubit.init([
             'cubit',
             '-nojournal',
@@ -113,7 +114,7 @@ def NWL_geom(
             '-information', 'off',
             '-warning', 'off',
             '-commandplugindir',
-            cubit_dir
+            str(cubit_dir)
         ])
     
     if logger == None or not logger.hasHandlers():
@@ -164,7 +165,9 @@ def NWL_geom(
     if source is not None:
         strengths = sm.source_mesh(vmec, source, export['dir'], logger = logger)
         
-        file = open('strengths.txt', 'w')
+        filepath = Path(export['dir']) / 'strengths.txt'
+
+        file = open(filepath, 'w')
         for tet in strengths:
             file.write(f'{tet}\n')
         
@@ -175,7 +178,7 @@ def extract_ss(ss_file):
     """Extracts list of source strengths for each tetrahedron from input file.
 
     Arguments:
-        ss_file (str): source strength input file.
+        ss_file (str): path to source strength input file.
 
     Returns:
         strengths (list): list of source strengths for each tetrahedron (1/s).
@@ -313,7 +316,7 @@ def find_coords(vmec, wall_s, phi, pt):
     return theta
 
 
-def flux_coords(vmec, wall_s, coords, coords_x, coords_y):
+def flux_coords(vmec, wall_s, coords):
     """Computes flux-coordinate toroidal and poloidal angles corresponding to
     specified Cartesian coordinates.
     
@@ -322,17 +325,13 @@ def flux_coords(vmec, wall_s, coords, coords_x, coords_y):
         wall_s (float): closed flux surface label extrapolation at wall.
         coords (array of array of float): Cartesian coordinates of all particle
             surface crossings (cm).
-        coords_x (array of float): x coordinates of all particle surface
-            crossings (cm).
-        coords_y (array of float): y coordinates of all particle surface
-            crossings (cm).
 
     Returns:
         phi_coords (array of float): toroidal angles of surface crossings (rad).
         theta_coords (array of float): poloidal angles of surface crossings
             (rad).
     """
-    phi_coords = np.arctan2(coords_y, coords_x)
+    phi_coords = np.arctan2(coords[:,1], coords[:,0])
     theta_coords = []
     
     for pt, phi in zip(coords, phi_coords):
@@ -352,28 +351,20 @@ def extract_coords(source_file):
     Returns:
         coords (array of array of float): Cartesian coordinates of all particle
             surface crossings.
-        coords_x (array of float): x coordinates of all particle surface
-            crossings (cm).
-        coords_y (array of float): y coordinates of all particle surface
-            crossings (cm).
     """
     import h5py
     
     # Load source file
     file = h5py.File(source_file, 'r')
     # Extract source information
-    dataset = file['source_bank']
-    # Extract Cartesian coordinates of particle crossings
-    coords_x = dataset['r']['x']
-    coords_y = dataset['r']['y']
-    coords_z = dataset['r']['z']
+    dataset = file['source_bank']['r']
     # Construct matrix of particle crossing coordinates
-    coords = np.empty((len(coords_x), 3))
-    coords[:,0] = coords_x
-    coords[:,1] = coords_y
-    coords[:,2] = coords_z
+    coords = np.empty((len(dataset), 3))
+    coords[:,0] = dataset['x']
+    coords[:,1] = dataset['y']
+    coords[:,2] = dataset['z']
 
-    return coords, coords_x, coords_y
+    return coords
 
 
 def plot(NWL_mat, phi_pts, theta_pts, num_levels):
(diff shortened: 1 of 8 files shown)


gonuke:
@connoramoreno - This needs a rebase before we can review it properly

NWL/NWL.py
```diff
@@ -0,0 +1,451 @@
...
+
+        for i, phi_bin in enumerate(phi_bins):
+            # Conditionally contribute to bin crossing count if crossing within
+            # bin
+            if np.abs(phi - phi_bin) <= tor_ext/(num_phi - 1)/2:
+                for j, theta_bin in enumerate(theta_bins):
+                    # Conditionally contribute to bin crossing count if
+                    # crossing within bin
+                    if np.abs(theta - theta_bin) <= pol_ext/(num_theta - 1)/2:
+                        count_mat[i,j] += 1
```
gonuke, line 435: Can we use Numpy's `histogram2d` for this? https://numpy.org/doc/stable/reference/generated/numpy.histogram2d.html

NWL/NWL.py
```diff
@@ -0,0 +1,451 @@
...
+
+    phi_bins = np.linspace(0.0, tor_ext, num = num_phi)
+    theta_bins = np.linspace(-pol_ext/2, pol_ext/2, num = num_theta)
+    
+    # Initialize count matrix
+    count_mat = np.zeros((num_phi, num_theta))
+
+    for pt in coords:
+        # Extract Cartesian coordinates and format as array
+        pt = [pt['x'], pt['y'], pt['z']]
```
gonuke, line 418: I wonder if there is a way to extract this info with h5py to get them in tuples of size three directly, rather than named like this?

NWL/NWL.py
```diff
@@ -0,0 +1,451 @@
...
+                    # crossing within bin
+                    if np.abs(theta - theta_bin) <= pol_ext/(num_theta - 1)/2:
+                        count_mat[i,j] += 1
+
+    # Define fusion neutron energy (eV)
+    n_energy = 14.1e6
+    # Define eV to joules constant
+    eV2J = 1.60218e-19
+    # Compute total neutron source strength (n/s)
+    strengths = extract_ss(ss_file)
```
gonuke, line 442: This raises a separate question about assuming that these source strengths are normalized to the total reactor power.  I wonder if total reactor power should be a separate input?

gonuke (changes requested):
A few comments on the novel code here, but I'll wait for the rebase to avoid re-reviewing other changes.

NWL/NWL.py
connoramoreno: We can of course add it as an input. Right now, the source strengths (and by extension fusion power) are as accurate as the plasma temperature and ion density profiles are.

gonuke: I only say this because for the purpose of OpenMC, they only need to describe the relative source strength - OpenMC will effectively normalize them upon reading them.  I think the pathway that we use to generate them in ParaStell currently ensures that they are meaningful on an absolute sense, but that may not always be true.

NWL/__init__.py
gonuke, line 1: I don't think we want/need this (empty) file

parastell.py
```diff
@@ -119,33 +128,33 @@ def legacy_export():
         
         # DAGMC export
         cubit.cmd(
-            f'export dagmc "dagmc.h5m" {facet_tol_str} {len_tol_str} '
+            f'export dagmc "{dir}{filename}.h5m" {facet_tol_str} {len_tol_str} '
```
gonuke, line 131: The best way to do this is use the python `Path` packages and form a fully qualified file path

parastell.py
```diff
@@ -184,17 +193,15 @@ def native_export():
         cubit.cmd("mesh surface all")
 
         # Export DAGMC file
-        cubit.cmd(f'export cf_dagmc "{cwd + "/dagmc.h5m"}" overwrite')
+        cubit.cmd(f'export cf_dagmc "{dir}{filename}.h5m" overwrite')
```
gonuke, line 196: See python `Path` package for more robust way to combine paths

parastell.py
```diff
@@ -184,17 +193,15 @@ def native_export():
...
 
-    # Get current working directory
-    cwd = os.getcwd()
+    dir = export['dir']
+    filename = export['h5m_filename']
 
-    # Import solids
     for name in components.keys():
-        cubit.cmd(f'import step "' + cwd + '/' + name + '.step" heal')
+        cubit.cmd(f'import step "{dir}{name}.step" heal')
```
gonuke, line 202: `Path` package

parastell.py
```diff
@@ -283,7 +306,7 @@ def exports(export, components, magnets, logger):
     if export['step_export']:
         logger.info('Exporting STEP files...')
         for name, comp in components.items():
-            cq.exporters.export(comp['solid'], name + '.step')
+            cq.exporters.export(comp['solid'], export['dir'] + name + '.step')
```
gonuke, line 309: python `Path` package

parastell.py
```diff
@@ -297,7 +320,9 @@ def exports(export, components, magnets, logger):
                 comp['solid'],
                 material_tags = [comp['h5m_tag']]
             )
-        model.export_dagmc_h5m_file()
+        model.export_dagmc_h5m_file(
+            filename = f"{export['dir']}{export['h5m_filename']}.h5m"
```
gonuke, line 324: python `Path` package

gonuke (changes requested):
Some new comments about combining paths.

NWL/NWL.py
connoramoreno: I couldn't find a clean way to do it via h5py directly but I think I came up with a decent alternative. New commit coming soon

magnet_coils.py
```diff
@@ -19,12 +23,9 @@ def mesh_magnets(vol_ids, logger):
...
     
-    # Assign export paths
-    cwd = os.getcwd()
-    base_name = 'coil_mesh'
-    general_export_path = f"{cwd}/{base_name}"
-    exo_path = f'{general_export_path}.exo'
-    h5m_path = f'{general_export_path}.h5m'
+    # Define export paths
+    exo_path = str(Path(export_dir) / 'coil_mesh.exo')
+    h5m_path = str(Path(export_dir) / 'coil_mesh.h5m')
```
gonuke, line 28: Do you need to convert to `str`? Have you tried without? If you change all path handling to use `pathlib` maybe this works fine?  Maybe paths are promoted to strings when necessary?  Just curious...

Either way, it could be easier to do eveything with pathlib until the last minute when you don't

gonuke: I noticed in L#541 below that you don't convert to string first...

parastell.py
```diff
@@ -184,17 +195,19 @@ def native_export():
...
-        cubit.cmd(f'export cf_dagmc "{cwd + "/dagmc.h5m"}" overwrite')
+        export_path = Path(dir) / f'{filename}.h5m'
+        cubit.cmd(
+            f'export cf_dagmc "{export_path}" overwrite'
+        )
 
-    # Get current working directory
-    cwd = os.getcwd()
+    dir = export['dir']
+    filename = export['h5m_filename']
```
gonuke, line 204: make these `Path` objects from the beginning?

parastell.py
```diff
@@ -280,10 +309,16 @@ def exports(export, components, magnets, logger):
             'Inclusion of magnets in H5M model requires Cubit export'
         )
     
+    dir = export['dir']
+    filename = export['h5m_filename']
```
gonuke, line 313: Initialize these as `Path` objects

gonuke (changes requested):
suggestions on streamlined use of `Path`

magnet_coils.py
connoramoreno: Cubit can accept `Path` objects but for some reason MOAB and CadQuery only accept strings

gonuke: Can you promote them to string at the last minute - as you pass them into the MOAB/CQ functions?  that will let you preserve them as Path objects everywhere else

parastell.py
connoramoreno: Regarding `filename`, since we combine it with the string `.h5m`, the syntax for compiling the whole export path would stay the same even if we convert it to a `Path` object. For example,
```
export_path = dir / f'{filename}.h5m'
```
where `dir` and `filename` are `Path` objects per your suggestion. In this case, I think it would make the most sense to keep `filename` as a string. However, if we have users supply `export['h5m_filename]` with the .h5m extension, e.g.,
```
export = {
    ...
    'h5m_filename': 'dagmc.h5m',
    ...
}
```
then converting `filename` to a `Path` object from the outset would make sense. What are your thoughts?

magnet_coils.py
connoramoreno: Certainly!

parastell.py
gonuke: or
```
export_path = dir / filename.with_suffix('.h5m')
```

gonuke (changes requested):
More thoughts on streamlining access to data

gonuke (approved):

PR #32, opened 2024-02-13
Skip imprinting

Edgar-21:
This PR skips the imprint step and instead merges the coincident surfaces using the fact that parastell will import the volumes in inside to outside order. (and some cubit geometry queries)

Likely we will want this to be some sort of setting, right now it just completely replaces the imprint/merge commands, maybe another item in the export dictionary would be the best place to put the option to toggle this, open to suggestions. 

For reference imprinting a build with 10 layers took about 50 minutes to imprint for me yesterday, this takes about 5 seconds to merge all coincident surfaces in the example script (7 surfaces to merge)


parastell.py
```diff
@@ -202,15 +204,80 @@ def native_export():
...
     for name in components.keys():
         import_path = dir / Path(name).with_suffix('.step')
         cubit.cmd(f'import step "{import_path}" heal')
         components[name]['vol_id'] = cubit.get_last_id("volume")
 
-    cubit.cmd('imprint volume all')
-    cubit.cmd('merge volume all')
-
+    merge_layer_surfaces()
+
```
gonuke, line 280: This should probably be an option that is triggered by some variable.

gonuke (changes requested):
I am not sure we want to permanently change this, but make it an option

parastell.py
```diff
@@ -202,15 +204,80 @@ def native_export():
...
+        #when using cubit.get_surface_type(int)
+        
+        # get a tuple of volume ids, these will be in the import order
+        # and are assumed to increase in order from inside to outside
+        volumes = cubit.get_entities('volume')
+
+        # tracks the surface id of the outer surface of the previous layer
+        lastOuterSurface = None
+        
+        for name, vol in zip(components.keys(),volumes):
```
gonuke, line 221: For robustness, you may want to build a dictionary of `components.keys()` to `volumes` when you build them rather than always relying on the ordering.  Something could come along in the future that disrupts the ordering and it will be hard to find.

parastell.py
```diff
@@ -202,15 +204,80 @@ def native_export():
...
+                for surf in surfaces:
+                    if cubit.get_surface_type(surf) == 'spline surface':
+                        splineSurfaces.append(surf)
+
+                if cubit.get_bounding_box('surface', splineSurfaces[1])[4] > cubit.get_bounding_box('surface',splineSurfaces[0])[4]:
+                    outerSurface = splineSurfaces[1]
+
+                else:
+                    outerSurface = splineSurfaces[0]
+
```
gonuke, line 247: Maybe this could be a function `orient_component_spoine_surfaces(vol)` that returns `[innerSurface, outerSurface]` and then use it here and in the next conditional?

gonuke: Or you do this operation once in both cases, and then act on it differently depending on which case.

gonuke (changes requested):
A few more coding suggetions.

parastell.py
Edgar-21: added it as an option in the `export` dict, updated example script. this will default to false.

parastell.py
Edgar-21: found that this already occurs in the cubit_export function, changed code to make use of that

parastell.py
Edgar-21: added local `orient_spline_surfaces` function to improve readability

magnet_coils.py
```diff
@@ -538,7 +538,7 @@ def magnet_coils(magnets, tor_ext, export_dir, logger = None):
         vol_ids = cut_mags(tor_ext, vol_ids, r_avg)
     
     # Export magnet coils
-    export_path = Path(export_dir) / Path(magnets["name"]).with_suffix('step')
+    export_path = Path(export_dir) / Path(magnets["name"]).with_suffix('.step')
```
gonuke, line 541: I think this '.' was removed in another PR somewhere

parastell.py
```diff
@@ -202,20 +203,92 @@ def native_export():
...
+                # the outer surface bounding box will have the larger max xy value
+                if cubit.get_bounding_box('surface', splineSurfaces[1])[4] > cubit.get_bounding_box('surface',splineSurfaces[0])[4]:
+                    outerSurface = splineSurfaces[1]
+                    innerSurface = splineSurfaces[0]
+
+                else:
+                    outerSurface = splineSurfaces[0]
+                    innerSurface = splineSurfaces[1]
+                
+            return innerSurface, outerSurface
```
gonuke, line 236: I don't think there's a good reason for this to be nested in the other function.  I think we should only do that when there is a driving reason to do so...

parastell.py
```diff
@@ -202,20 +203,92 @@ def native_export():
...
+
+            # get volume ID for layer
+
+            vol_id = components[name]['vol_id']
+
+            # wait to merge until the next layer if the plasma is included
+            # store surface to be merged for next loop
+            if name == 'plasma':
+                
+                _ , lastOuterSurface = orient_spline_surfaces(vol_id)
```
gonuke, line 252: Why not call this before the conditional - you need it in every branch

.gitignore
```diff
@@ -136,3 +136,7 @@ dmypy.json
 *.jou
 *.ods
 *.vtk
+*.cub
+*.cub5
+*.h5
+logMem.sh
```
gonuke, line 142: where does this file come from?

gonuke (changes requested):
A few more suggestions - nearly there...

.gitignore
Edgar-21: the cubit files and h5 are from testing different things, I suppose I should just delete these items out of my parastell directory rather than cluttering the gitignore?

logMem.sh definitely shouldn't be in here

magnet_coils.py
Edgar-21: without the '.' python complains about an invalid suffix, adding the '.' results in the magnets step file being exported, at least for me

parastell.py
Edgar-21: Should I bump `legacy_export` and `native_export` out to be their own functions as well then? Or is it ok to have these local to `cubit_export`?  I appreciate input on good style/practices since I don't have much sense of those. I had originally nested all these functions to level where they would get used to limit there scope

magnet_coils.py
gonuke: I must have misremembered that change - maybe it was making this same fix as this

parastell.py
gonuke: The export ones make slightly more sense to me than this one.  Not really sure why.  I guess there is no use case for calling the export functions from outside the `cubit_export` function, but I can imagine that this function could have such a use case.

.gitignore
gonuke: I was specifically asking about the `logMem.sh` file.  It seemed a little specific.

parastell.py
```diff
@@ -203,19 +204,89 @@ def native_export():
...
+        """
+        surfaces = cubit.get_relatives('volume', volume_id, 'surface')
+            
+        splineSurfaces = []
+
+        for surf in surfaces:
+            if cubit.get_surface_type(surf) == 'spline surface':
+                splineSurfaces.append(surf)
+
+        # check if this is the plasma
```
gonuke, line 218: In the future, there could be other reasons that there is only one surface - perhaps change this comment

gonuke (changes requested):
One little request to tweak a comment and we should be ready to go

gonuke:
I'm happy to approve this, but wonder if someone who can actively test it should be responsible for merging. (@connoramoreno @FusionSandwich ?)

Edgar-21:
Made some changes to comments, variable names, line lengths and newlines based on what I read in PEP8, please let me know if I missed anything

gonuke:
> Made some changes to comments, variable names, line lengths and newlines based on what I read in PEP8, please let me know if I missed anything

You should look into a PEP8 formatter for your editor that will do this automatically...

PR #33, opened 2024-02-13
added cubit reset

Edgar-21:
Looping Parastell to Make Different Geometry Results in All Previous Geometry Being Included in Each Model

Examples:
First iteration
image
9 layers visible (excluding plasma)
Second iteration
image
The only difference between these two should be the thickness of the layers, but it can be seen that there are many more layers due to both sets of step files existing in cubit at the time of the export.

Adding `cubit.cmd('reset')` at the very end of `parastell.parastell` resolves this, I put it here rather than in the `cubit_export` in case some other cubit operations ever need to take place after `cubit_export`. As always, open to suggestions

### DIFF
--- a/parastell.py
+++ b/parastell.py
@@ -861,3 +861,6 @@ def parastell(
             vmec, source, export['dir'], logger = logger
         )
         return strengths
+    
+    # reset cubit to avoid issues when looping
+    cubit.cmd('reset')
--- a/parastell.py
+++ b/parastell.py
@@ -847,7 +847,7 @@ def parastell(
 
     if magnets is not None:
         magnets['vol_id'] = magnet_coils.magnet_coils(
-            magnets, (repeat + 1)*tor_ext, export['dir'], logger = logger
+            magnets, (repeat + 1)*tor_ext, export_dict['dir'], logger = logger
         )
 
     try:
@@ -858,9 +858,10 @@ def parastell(
     
     if source is not None:
         strengths = source_mesh.source_mesh(
-            vmec, source, export['dir'], logger = logger
+            vmec, source, export_dict['dir'], logger = logger
         )
         return strengths
     
     # reset cubit to avoid issues when looping
-    cubit.cmd('reset')
+    if export_dict['h5m_export'] == 'Cubit':    
+        cubit.cmd('reset')


Edgar-21:
there is something else going on as well when looping, the logger logs each message x the iteration of the loop and it seems like memory utilization increases with each loop iteration, maybe something with cadquery? Does it need to get 'reset' like cubit? I haven't been able to find anything obvious in the documentation. Should I close this PR and make this an issue?

example of the logger behavior (on the 10th iteration):

```
10:35:12: Building plasma...
10:35:12: Building plasma...
10:35:12: Building plasma...
10:35:12: Building plasma...
10:35:12: Building plasma...
10:35:12: Building plasma...
10:35:12: Building plasma...
10:35:12: Building plasma...
10:35:12: Building plasma...
10:35:12: Building plasma...
10:35:30: Building sol...
10:35:30: Building sol...
10:35:30: Building sol...
10:35:30: Building sol...
10:35:30: Building sol...
10:35:30: Building sol...
10:35:30: Building sol...
10:35:30: Building sol...
10:35:30: Building sol...
10:35:30: Building sol...
```

connoramoreno:
I believe the logger behavior has to do with a new logger being instantiated with each loop iteration. As of right now, it doesn't seem like ParaStell is recognizing that there is already a logger in place (from the first loop iteration). I wonder if there's a way to have ParaStell recognize whether there's a logger without one being supplied as an argument

gonuke:
Don't you pass the logger to parastell? so it should reuse the same one each time if you pass the same one

connoramoreno:
If there isn't a logger passed to ParaStell, it will instantiate its own. In a loop where no logger is passed, ParaStell will instantiate a new logger with each iteration, and each logger will report the same thing as seen above. That's my best understanding of what's going on, at least

Edgar-21:
Parastell creates a logger if one isn't passed which is what I believe causes the repetitive logging behavior in a loop, passing a logger does fix this behavior, thanks @connoramoreno. I think there is still some sort of memory issue when iterating. I'm working on collecting some data but it looks like generally speaking on each iteration the free memory will decrease until ultimately it runs out while building the layers in CadQuery. I can resume the script with the same model that killed the processes previously and it will run no problem. Example output for when the process is killed

```
16:17:48: New stellarator build
16:17:48: Building plasma...
16:18:05: Building sol...
Killed
```
Here is plot of the free memory from the last run of this script
image
Not really a clear cut trend. This data isn't perfect since I was using my machine for other things during portions of this run, looks like it gets close to zero free memory in early iterations however. I'll be running this again tonight I expect so hopefully can get some cleaner data.

Edgar-21:
Not sure what is causing the memory issue, looping parastell using only cadquery doesn't seem to exhibit the same behavior, might be something for another day. I updated the cubit reset call to be conditional on if cubit is being used for the h5m export.

Edgar-21:
I'm just going to add these changes to #32 (cubit reset, fixes to variable names)

PR #34, opened 2024-02-13
Export magnet tet mesh via Cubit

connoramoreno:
Changes the Cubit export of the magnet tetrahedral mesh directly to H5M format, rather than a Cubit export in Exodus format and subsequent conversion to H5M via MOAB. Note that direct export to H5M via Cubit requires Cubit version 2023.11.

### DIFF
--- a/magnet_coils.py
+++ b/magnet_coils.py
@@ -1,6 +1,5 @@
 import log
 import cubit
-from pymoab import core, types
 import numpy as np
 from sklearn.preprocessing import normalize
 import os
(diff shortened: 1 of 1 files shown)


magnet_coils.py
```diff
@@ -23,18 +22,10 @@ def mesh_magnets(vol_ids, export_dir, logger):
...
-    # EXODUS export
-    cubit.cmd(f'export mesh "{exo_path}"')
-    
-    # Convert EXODUS to H5M
-    mb = core.Core()
-    exodus_set = mb.create_meshset()
-    mb.load_file(str(exo_path), exodus_set)
-    mb.write_file(str(h5m_path), [exodus_set])
+    # H5M export
+    cubit.cmd(f'export cf_dagmc "{export_path}" overwrite')
```
gonuke, line 28: I'm curious about the workflow here. Historically, a `dagmc` export was only for surface triangles and never volume mesh.  Is that changed for `cf_dagmc`?  Does this export both surface triangle info in a DAGMC-usable form as well as volume tetmesh?  Or just the mesh? How does it know?

gonuke (commented):
A question to clarify the workflow a little

magnet_coils.py
connoramoreno: You're right, only the surface triangles get exported and not the full tetrahedra. I hadn't thought to look at the insides of the resultant mesh

PR #35, opened 2024-02-13
Fix typo in magnet STEP export

connoramoreno:
Very small PR to fix a typo in the magnet STEP export. Accidentally omitted a "." in the magnet STEP file export path.


gonuke (approved):
More evidence that we need testing :)

PR #41, opened 2024-03-04
Magnets oo

### LINKED ISSUE #43: First pass to OO for magnets

Edgar-21:
looking for feedback on testing/overall flow, and if docstrings are sufficient. much of the functionality is copied from the old manget_coils.py

closes #43


tests/test_magnet_coils.py
```diff
@@ -0,0 +1,150 @@
...
+
+toroidal_extent = 90
+
+export_dir = ''
+
+
+def test_rectangular_magnets():
+
+    magnets = {
+    'file': 'coils.txt',
```
gonuke, line 32: can you make sure to add this test file

src/magnet_coils.py
```diff
@@ -0,0 +1,534 @@
...
+
+        Arguments:
+            vec (array of [float, float, float]): input vector to be normalized.
+
+        Returns:
+            vec (array of [float, float, float]): normalized input vector.
+        """
+        vec = normalize(vec.reshape(-1, 1), axis=0).ravel()
+
+        return vec
```
gonuke, line 389: We should figure out how to make this available to all of parastell instead of in this class.

gonuke (changes requested):
Thanks @Edgar-21 - a few high level comments/requests.

I'm holding back on larger redesign/refactor questions in order to get this in place ASAP, but I do have some thoughts on how to make changes things a little more streamlined.

src/magnet_coils.py
Edgar-21: ill pop this out to be its own function that can get relocated and imported as desired once we figure out the structure

src/magnet_coils.py
```diff
@@ -0,0 +1,534 @@
...
+
+        if logger == None or not logger.hasHandlers():
+            self.logger = log.init()
+        else:
+            self.logger = logger
+
+        self.extract_filaments()
+        self.set_average_radial_distance()
+        self.extract_cross_section()
+        self.set_filtered_filaments()
```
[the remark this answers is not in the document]
Edgar-21: didn't make changes to this for now

src/magnet_coils.py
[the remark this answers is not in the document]
Edgar-21: added but not tested, pending yaml file

src/magnet_coils.py
```diff
@@ -0,0 +1,582 @@
...
+def unit_vector(vec):
+    """Normalizes given vector.
+
+    Arguments:
+        vec (array of [float, float, float]): input vector to be normalized.
+
+    Returns:
+        vec (array of [float, float, float]): normalized input vector.
+    """
+    vec = normalize(vec.reshape(-1, 1), axis=0).ravel()
```
gonuke, line 21: I wonder if we can do this without an extra dependency on `sklearn` for this?  We put something together based on only `numpy` in the slice that @connoramoreno is working on.

gonuke (changes requested):
Assuming your tests are passing, this looks to be nearly ready...

src/magnet_coils.py
Edgar-21: yeah it can be done with numpy

Edgar-21:
tests are passing

src/magnet_coils.py
```diff
@@ -0,0 +1,582 @@
...
+    """Normalizes given vector.
+
+    Arguments:
+        vec (array of [float, float, float]): input vector to be normalized.
+
+    Returns:
+        vec (array of [float, float, float]): normalized input vector.
+    """
+
+    vec = vec/((np.sum(vec**2))**0.5)
```
gonuke, line 21: `np.squared` and `np.sqrt` may be safer.  Have we tested that this works with an array of vectors?

gonuke (changes requested):
best practice for numpy based normalize

src/magnet_coils.py
Edgar-21: `np.square` definitely would be a good call. np.sum() should always return a scalar in our use case (normalizing 1D arrays) so I would imagine the **0.5 should be fine but maybe it would be best to just use np.sqrt anyway? I could add a check as well to make sure the results are as expected? Either in the function itself or as a test?

gonuke: I don't think we need to add a test here, but it would be good to test this in a notebook just to make sure it works in the case with an Numpy array of N vectors - kind of like we were doing on Saturday.

gonuke: Also, I always have the sense that `sqrt()` functions are more robust that `**0.5`, and the `np.sum` will be a vector of scalars if `vec` is effectively a list of vectors.

gonuke: There is a version of such a function in #55 and we should probably pick one and be confident it will work in all our use cases and make it available to all the methods. (ping @connoramoreno )

connoramoreno: I've added a `utils.py` utility script to the `src` directory that contains the `normalize()` function I use in #55, though that `normalize()` function only works on 2-D arrays. I can change it to work for arrays of any dimension but it might be better just to use the `normalize()` function from the `scikit-learn.preprocessing` Python module, as that has all the functionality we'd need.

gonuke: my first instinct was to avoid another dependency for just this functionality, but I'm happy to hear arguments about one vs the other vis-a-vis robustness

connoramoreno: Well we know for certain that `sklearn.preprocessing.normalize()` is robust and can handle arrays of any dimension, with some caveats for 1-D arrays (though maybe that isn't important for our use case). I suppose `src.utils.normalize()` as-is can also handle ndarrays but not 1-D arrays (again, may not be important for our use case)

Edgar-21: `np.linalg.norm` might be another candidate? Would need to handle 1D arrays specially (axis 1 doesnt exist) but can be done with numpy. Example:

```
array([[1, 2, 3],
       [4, 5, 6],
       [7, 8, 9]])
>>> norm_vec = np.linalg.norm(arr,axis=1)
>>> norm_vec
array([ 3.74165739,  8.77496439, 13.92838828])
>>> normalized_arr = np.divide(arr.T, norm_vec).T
>>> normalized_arr
array([[0.26726124, 0.53452248, 0.80178373],
       [0.45584231, 0.56980288, 0.68376346],
       [0.50257071, 0.57436653, 0.64616234]])
>>> np.linalg.norm(normalized_arr, axis=1)
array([1., 1., 1.])

>>> arr = np.array([1,2,3])
>>> norm_vec = np.linalg.norm(arr, axis=1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/filespace/e/epflug/research/mambaforge/lib/python3.10/site-packages/numpy/linalg/linalg.py", line 2583, in norm
    return sqrt(add.reduce(s, axis=axis, keepdims=keepdims))
numpy.exceptions.AxisError: axis 1 is out of bounds for array of dimension 1
>>> 


```

gonuke (approved):

PR #55, opened 2024-03-05
OO version of in-vessel component geometry construction

### LINKED ISSUE #44: First pass to OO for blanket geometry
Convert blanket geometry building code to object oriented with minimal redesign.

connoramoreno:
Creates an object-oriented version of in-vessel component (IVC) geometry construction. The new Stellarator class features three primary child classes:

- Data class: contains all data for IVCs, including outer surface Cartesian point-loci
- CADGeometry class: contains all CAD information for IVCs, including the corresponding volume IDs in Cubit-generated H5M file. This class also contains the magnet geometry child class MagnetSet
- SourceMesh class: contains the source mesh H5M file

Passes all included tests.

Fixes #44.


gonuke:
High level design questions:
* I think it makes sense to have this build only the invessel components and completely decouple the magnet and source mesh generation.  
* If so, then I think we also move all the export to somewhere else.

The final design is: 
* parastell
    * reads input
    * initializes cad engines as required, esp cubit
    * calls for each requested component to do its work
        *  ivc
        * magnets
        * source mesh
    * exports data as requested

gonuke:
The follow up question is whether we do that before merging this or after?

connoramoreno:
I think that makes sense! I don't think it would take long to restructure the script to accommodate that.

connoramoreno:
I've separated building the IVCs from the main script as suggested. It's worth noting though that `pytest` fails even though the scripts work. Specifically, running the test scripts using Python doesn't result in any errors and everything runs smoothly. When using `pytest` however, there's a library linking issue related to Cubit:

```
___________________________________________________________________________________________________ ERROR collecting test_parastell.py ____________________________________________________________________________________________________
ImportError while importing test module '/home/camoreno/parastell/tests/test_parastell.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../miniforge3/envs/ps_env/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
test_parastell.py:1: in <module>
    import src.parastell as ps
../src/parastell.py:1: in <module>
    import cubit
/opt/Coreform-Cubit-2023.11/bin/cubit.py:44: in <module>
    from cubit3 import *
/opt/Coreform-Cubit-2023.11/bin/cubit3.py:12: in <module>
    import _cubit3
E   ImportError: /opt/Coreform-Cubit-2023.11/bin/libshowviz_base.so: undefined symbol: _ZN22vtkXOpenGLRenderWindow11SetWindowIdEm
```

I tried for a while to fix this but I'm really not sure what's going on.

gonuke (changes requested):
I got started with a few comments here, but as I got deeper, I wanted to think about how to simplify this a little.  I put some ideas here: https://docs.google.com/presentation/d/10_hmTwkoqR1gxFOxLZhh6XAQQXSNU8qs8qN-A0s_Htg/edit#slide=id.g2c0036363df_0_9

src/ivc.py
```diff
@@ -0,0 +1,656 @@
...
+
+    def generate_data(self):
+        '''Generates data defining stellarator in-vessel component geometry.
+        '''
+        n_phi = len(self.phi_list)
+        n_theta = len(self.theta_list)
+
+        # Extract toroidal extent of build
+        self.seg_tor_ext = self.phi_list[-1]
+        self.tot_tor_ext = (self.repeat + 1)*self.seg_tor_ext
```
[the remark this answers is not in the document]
connoramoreno: The toroidal extent is also defined independently for the source mesh. I found that keeping the geometry and source toroidal extent definitions independent was useful since one may want to repeat the segment geometry in OpenMC (or other MC code) rather than ParaStell for improved performance. In this case, the source still needs to span the full geometry used in MC, which would be different from the geometry modeled in ParaStell

connoramoreno:
Unfortunately, I realized too late that I forgot to commit renaming the files after doing `git mv` so I lost the git history from ivc.py and test_ivc.py. Luckily, there wasn't much history to begin with.

connoramoreno:
Also unfortunately, the same issue persists with `pytest` (though I didn't expect that to change with the updated structure).

connoramoreno:
I think this updated structure is much cleaner by the way; the in-vessel component script is far more readable, not to mention significantly shorter. Thanks, @gonuke!

src/invessel_build.py
```diff
@@ -0,0 +1,480 @@
+import argparse
+import log
+import read_vmec
+import cadquery as cq
+import numpy as np
+import math
+from scipy.interpolate import RegularGridInterpolator
+import yaml
+
+m2cm = 100
```
gonuke, line 10: Eventually, we should define this in one place and reuse it.

src/invessel_build.py
```diff
@@ -0,0 +1,480 @@
...
+            logger=None
+    ):
+        self.vmec = vmec
+        self.build = build
+        self.repeat = repeat
+        self.num_phi = num_phi
+        self.num_theta = num_theta
+        self.scale = scale
+        self.plasma_h5m_tag = plasma_h5m_tag
+        self.sol_h5m_tag = sol_h5m_tag
```
gonuke, line 119: Let's initialize these here

```suggestion
        self.sol_h5m_tag = sol_h5m_tag

        self.Surfaces = []
        self.Components = []

```

src/invessel_build.py
```diff
@@ -0,0 +1,480 @@
...
+        }
+
+        self.phi_list = expand_ang(self.build['phi_list'], self.num_phi)
+        self.theta_list = expand_ang(self.build['theta_list'], self.num_theta)
+
+    def populate_surfaces(self):
+        '''Populates Surface class objects representing the outer surface of
+        each component specified in the radial build.
+        '''
+        self.Surfaces = []
```
gonuke, line 165: Delete this if it is initialized in the constructor above

src/invessel_build.py
```diff
@@ -0,0 +1,480 @@
...
+                self.logger.error(e.args[0])
+                raise e
+
+            if name == 'plasma':
+                s = 1.0
+            else:
+                s = self.build['wall_s']
+
+            if 'h5m_tag' not in layer_data:
+                layer_data['h5m_tag'] = name
```
gonuke, line 187: I think the above logic of the defaults means that the `h5m_tag` for plasma and sol could be `Null`. Is that OK? Or do we need to test inputs to prevent that.

src/invessel_build.py
```diff
@@ -0,0 +1,480 @@
...
+        [surface.calculate_loci() for surface in self.Surfaces]
+
+    def generate_components(self):
+        '''Constructs a CAD solid for each component specified in the radial
+        build by cutting the interior surface solid from the outer surface
+        solid for a given component.
+        '''
+        self.logger.info(f'Constructing in-vessel components...')
+
+        self.Components = []
```
gonuke, line 235: Remove this if initialized above

src/parastell.py
```diff
@@ -1,150 +1,647 @@
...
+                self.magnets is not None):
+            raise ValueError(
+                'Inclusion of magnets in H5M model requires Cubit export'
+            )
+    
+    def init_cubit(self):
+        '''Initializes Coreform Cubit with the DAGMC plugin.
+        '''
+        cubit_dir = os.path.dirname(inspect.getfile(cubit))
+        cubit_dir = Path(cubit_dir) / Path('plugins')
```
gonuke, line 238: Maybe call this `cubit_plugin_dir` for clarity

src/parastell.py
```diff
@@ -1,150 +1,647 @@
...
+        self.invessel_build = ivb.invessel_build
+        self.invessel_build.populate_surfaces()
+        self.invessel_build.calculate_loci()
+        self.invessel_build.generate_components()
 
-        self.rib = cq.Workplane("XY").spline(self.r_loci).close()
+    def construct_source_mesh(self):
+        '''Constructs SourceMesh class object.
+        '''
+        self.source_mesh = SourceMesh(self.vmec, self.source)
```
gonuke, line 265: Because of how you've imported `source_mesh` doesn't this need to be
```suggestion
        self.source_mesh = source_mesh.SourceMesh(self.vmec, self.source)
```

src/parastell.py
```diff
@@ -1,150 +1,647 @@
...
-        self.r_loci = np.append(self.r_loci, self.r_loci[0,:])
+    def construct_invessel_build(self):
+        '''Construct InVesselBuild class object.
+        '''
+        ivb = InVesselBuild(
+            self.vmec, self.build, self.repeat, self.num_phi, self.num_theta,
+            self.scale, self.export_dict['plasma_h5m_tag'],
+            self.export_dict['sol_h5m_tag'], self.logger
+        )
+        self.invessel_build = ivb.invessel_build
```
gonuke, line 257: You import `invessel_build` as `ivb` so I think you are overwriting this variable here.  Instead, let's set it to be the member variable right away. 

```suggestion
        self.invessel_build = InVesselBuild(
            self.vmec, self.build, self.repeat, self.num_phi, self.num_theta,
            self.scale, self.export_dict['plasma_h5m_tag'],
            self.export_dict['sol_h5m_tag'], self.logger
        )
```

src/parastell.py
```diff
@@ -1,150 +1,647 @@
...
+        export_path = (
+            Path(self.export_dict['dir']) /
+            Path(self.export_dict['h5m_filename']).with_suffix('.h5m')
+        )
+        model.export_dagmc_h5m_file(
+            filename=export_path
+        )
+
+
+class InVesselBuild(object):
```
gonuke, line 520: I don't think we need this class, right?

src/parastell.py
```diff
@@ -1,150 +1,647 @@
...
+            scale,
+            plasma_h5m_tag,
+            sol_h5m_tag,
+            logger
+    ):
+        self.invessel_build = ivb.InVesselBuild(
+            vmec, build, repeat, num_phi, num_theta, scale, plasma_h5m_tag, sol_h5m_tag, logger)
+
+
+class MagnetSet(object):
```
gonuke, line 561: We shouldn't need this class

src/parastell.py
```diff
@@ -1,150 +1,647 @@
...
+        magnets,
+        tor_ext,
+        export_dir,
+        logger
+    ):
+        """self.magnet_geometry = magnet_coils.MagnetSet(
+            magnets, tor_ext, export_dir, logger)"""
+
+
+class SourceMesh(object):
```
gonuke, line 584: We shouldn't need this class

src/parastell.py
```diff
@@ -1,150 +1,647 @@
...
+        '''
+        self.source_mesh = SourceMesh(self.vmec, self.source)
 
-    def vmec2xyz(self, poloidal_offset=0):
+    def construct_magnets(self):
+        '''Constructs MagnetSet class object.
         '''
-        Return an N x 3 NumPy array containing the Cartesian coordinates of the points
-        at this toroidal angle and N different poloidal angles, each offset slightly
+        self.magnet_set = MagnetSet(
```
gonuke, line 270:
```suggestion
        self.magnet_set = magnet_coils.MagnetSet(
```

gonuke (changes requested):
This does look a lot cleaner (if I do say so myself).  A few more comments and suggestions here.  Probably will need one more round of review after I see how these changes settle in.

src/invessel_build.py
```diff
@@ -0,0 +1,480 @@
...
+
+        tangent = next_pt_loci - self.rib_loci
+
+        plane_norm = np.array([-np.sin(self.phi), np.cos(self.phi), 0])
+
+        norm = np.cross(plane_norm, tangent)
+
+        return self.normalize(norm)
+
+    def normalize(self, vec_list):
```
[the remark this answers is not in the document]
connoramoreno: I've created a utility script (`utils.py`) to house these kinds of utility functions. So far, I've put `normalize()` and `expand_ang()` in the script, but we can discuss this structure.

src/invessel_build.py
connoramoreno: Good point. The intended workflow is that if `plasma_h5m_tag` and `sol_h5m_tag` are `None`, the applied material tag for each will be set to the respective name (i.e., the tag for the plasma will be `'plasma'` and that for the SOL will be `'sol'`) but it seems I accidentally dropped that functionality during my change to OO. I'll rework the script to make sure that functionality is restored, unless you have a suggestion for an alternative workflow.

gonuke (changes requested):
A couple of final little cleanup/changes

gonuke:
Since I just merged #41 that has a copy of `src/utils.py` this probably needs a rebase

src/invessel_build.py
```diff
@@ -0,0 +1,452 @@
...
+            if self.sol_h5m_tag:
+                self.build['radial_build']['sol']['h5m_tag'] = self.sol_h5m_tag
+
+        self.build['radial_build'] = {
+            'plasma': {
+                'thickness_matrix': np.zeros((
+                    len(self.build['phi_list']),
+                    len(self.build['theta_list'])
+                )),
+                'h5m_tag': self.plasma_h5m_tag
```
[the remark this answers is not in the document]
connoramoreno: Yes, my bad

gonuke (approved):
A few more changes... I think the export part is difficult to follow and a little scattered.  I'll be proposing some new design for that, probably.

gonuke (changes requested):
I meant to request changes and not approve the last review.

src/sourcemesh.py
```diff
@@ -282,50 +292,52 @@ def _create_tets_from_hex(self, s_idx, theta_idx, phi_idx):
...
-            [0, 0            ,  0],
-            [0, theta_idx    ,  0],
+            [0, 0,  0],
+            [0, theta_idx,  0],
             [0, theta_idx + 1,  0],
-            [0, 0            ,  1],
-            [0, theta_idx    ,  1],
+            [0, 0,  1],
+            [0, theta_idx,  1],
             [0, theta_idx + 1,  1]
```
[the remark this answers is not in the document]
connoramoreno: Yes, definitely more readable as it was; my change wasn't intentional, just blindly using an auto-formatter

src/parastell.py
```diff
@@ -1,150 +1,602 @@
...
-        self.rib = cq.Workplane("XY").spline(self.r_loci).close()
+    def construct_source_mesh(self):
+        """Constructs SourceMesh class object.
+        """
+        self.source_mesh = sm.SourceMesh(
+            self.vmec, self.source['num_s'], self.source['num_theta'],
+            self.source['num_phi'], self.source['tor_ext'], self.logger
+        )
+        self.source_mesh.create_vertices()
+        self.source_mesh.create_mesh()
```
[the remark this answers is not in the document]
connoramoreno: `strengths` is a member variable of the `SourceMesh` class, so it is available as part of that class object. We could alternatively make `strengths` a member variable of the `Stellarator` class after `SourceMesh` is constructed. As is, `strengths` can be accessed via
``` python
stellarator = parastell.Stellarator(...)
stellarator.construct_source_mesh(...)
strengths = stellarator.source_mesh.strengths
```

src/parastell.py
```diff
@@ -1,150 +1,602 @@
...
+        self.components = {}
 
-        tangent = next_pt_loci - self.r_loci
+        for component, (name, layer_data) in zip(
+            self.invessel_build.Components,
+            self.build['radial_build'].items()
+        ):
+            self.components[name] = {}
+            self.components[name]['h5m_tag'] = layer_data['h5m_tag']
+            self.components[name]['solid'] = component
```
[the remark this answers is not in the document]
connoramoreno: The method essentially combines the in-vessel component and magnet CAD information into a single dictionary but this data structure definitely isn't necessary for the export procedure. I constructed it this way to make component information (volume IDs and material tags) centrally available for the user after CAD construction, meshing, etc.

I figured if the data is going to be structured this way already, I might as well use it, although I agree it might not be the most streamlined methodology.

We can alternatively have the user access this information from separate `InVesselBuild` and `MagnetSet` class objects on the backend.

connoramoreno:
This PR has become rather large, sorry.

src/cubit_utils.py
```diff
@@ -0,0 +1,294 @@
...
+
+    Arguments:
+        cubit_flag (bool): flag to indicate whether Coreform Cubit has been
+            initialized.
+
+    Returns:
+        cubit_flag (bool): flag to indicate whether Coreform Cubit has been
+            initialized.
+    """
+    if not cubit_flag:
```
gonuke, line 19: More meaningful name here
```suggestion
    if not cubit_initialized:
```

src/cubit_utils.py
```diff
@@ -0,0 +1,294 @@
...
+            '-warning', 'off',
+            '-commandplugindir',
+            str(cubit_plugin_dir)
+        ])
+        cubit_flag = True
+
+    return cubit_flag
+
+
+def import_step_cubit(filename, import_dir):
```
gonuke, line 37: should we call `init_cubit` in each of these functions?  After the first time, it will basically do nothing, but at least we'll always know it's been initialized.

src/cubit_utils.py
```diff
@@ -0,0 +1,294 @@
...
+
+    Arguments:
+        volume_ids (iterable of int): iterable of Cubit volume IDs.
+        filename (str): name of H5M output file, excluding '.h5m' extension.
+        export_dir (str): directory to which to export the H5M output file
+            (defaults to empty string).
+    """
+    for vol in volume_ids:
+        cubit.cmd(f'volume {vol} scheme tetmesh')
+        cubit.cmd(f'mesh volume {vol}')
```
gonuke, line 78: Separation of concerns: Should we separate mesh generation from mesh export?

src/cubit_utils.py
```diff
@@ -0,0 +1,294 @@
...
+
+    exo_path = Path(export_dir) / Path(filename).with_suffix('.exo')
+    h5m_path = Path(export_dir) / Path(filename).with_suffix('.h5m')
+
+    cubit.cmd(f'export mesh "{exo_path}" overwrite')
+    subprocess.run(f'mbconvert {exo_path} {h5m_path}', shell=True)
+    Path.unlink(exo_path)
+
+
+def orient_spline_surfaces(volume_id):
```
gonuke, line 88: I think this could live with the IVC stuff.  I was imagining these utils being largely (entirely) about import/export and not about geometry querying/modification.

Similarly for `merge_layer_surface` but happy to discuss.

src/cubit_utils.py
```diff
@@ -0,0 +1,294 @@
...
+    facet_tol_str = ''
+    len_tol_str = ''
+    norm_tol_str = ''
+
+    if faceting_tolerance is not None:
+        facet_tol_str = f'faceting_tolerance {faceting_tolerance}'
+    if length_tolerance is not None:
+        len_tol_str = f'length_tolerance {length_tolerance}'
+    if normal_tolerance is not None:
+        norm_tol_str = f'normal_tolerance {normal_tolerance}'
```
gonuke, line 221:
```suggestion
    tol_str = ''

    if faceting_tolerance is not None:
        tol_str += f'faceting_tolerance {faceting_tolerance} '
    if length_tolerance is not None:
        tol_str += f'length_tolerance {length_tolerance} '
    if normal_tolerance is not None:
        tol_str += f'normal_tolerance {normal_tolerance}'
```

src/cubit_utils.py
```diff
@@ -0,0 +1,294 @@
...
+
+    export_path = Path(export_dir) / Path(filename).with_suffix('.h5m')
+    cubit.cmd(
+        f'export dagmc "{export_path}" {facet_tol_str} {len_tol_str} '
+        f'{norm_tol_str} make_watertight'
```
gonuke, line 226:
```suggestion
        f'export dagmc "{export_path}" {tol_str} make_watertight'
```

src/cubit_utils.py
```diff
@@ -0,0 +1,294 @@
...
+            (defaults to 'dagmc').
+        export_dir (str): directory to which to export the DAGMC output file
+            (defaults to empty string).
+    """
+    if skip_imprint:
+        merge_layer_surfaces(components_dict)
+
+    else:
+        cubit.cmd('imprint volume all')
+        cubit.cmd('merge volume all')
```
gonuke, line 200: Put this in a little function and share it with the `native` method

gonuke (changes requested):
Some changes to the cubit utils and then I'll think we'll merge and revise after merging as needed

src/cubit_utils.py
connoramoreno: Separating functionality the way you suggest makes sense to me, but I wonder how this separation of concerns affects the other utility functions.

E.g., with tetrahedral mesh generation (mentioned above) should just the mesh export live within `cubit_utils.py` while the mesh generation lives within `magnet_coils.py`? Similarly, for DAGMC exports, should imprinting/merging and material tagging live within `parastell.py` while `cubit_utils.py` handles only the export?

The legacy and native DAGMC exports handle material tagging and surface faceting differently, so my concern is that this will lead to many rather small functions specific to export methods. That may be fine but I worry it might be overcomplicating things.

connoramoreno: Also, this routine that skips imprinting depends on STEP files being imported to Cubit, which is called by `parastell.py` and updates component information to include volume IDs for material tagging in Cubit. Because that information belongs to `parastell.py` rather than `invessel_build.py`, I wonder if it might be better to put `merge_layer_surface` and `orient_spline_surfaces` in `parastell.py` rather than `invessel_build.py` as suggested.

src/cubit_utils.py
connoramoreno: For the Cubit utility script, I only included `init_cubit()` in `import_step_cubit()` since, presumably, if the export functions are being called, Cubit has already been initialized.

However, `init_cubit()` is included in all functions with Cubit functionality outside `cubit_utils.py` that don't rely on Cubit operations already happening.

connoramoreno:
I'd still like to discuss the implementation of default dictionaries before merging this PR

gonuke (approved):
Thanks for sticking with this @connoramoreno

PR #56, opened 2024-03-06
Yamlification of parastell

### LINKED ISSUE #45: Develop initial method for reading input from YAML
Define structure of YAML file and build standard method for reading and parsing yaml.

FusionSandwich:
Included Yaml input file and new python execution file. Current issue with getting to write to a yaml output file. Will likely be updated depending on how strengths output changes.

Closes #45

### DIFF
--- a/example.py
+++ b/example.py
@@ -1,0 +1,51 @@
+import yaml
+import parastell
+import numpy as np
+
+# Load configuration from YAML file
+config_file = 'config.yaml'
+with open(config_file, 'r') as file:
+    config = yaml.safe_load(file)
+
+# Access radial_build configurations
+radial_build = config['radial_build']
+phi_list = radial_build['phi_list']
+theta_list = radial_build['theta_list']
+repeat = radial_build['repeated'] 
+wall_s = radial_build['wall_s'] 
+
+# Correctly access stellarator_parameters
+stellarator_parameters = config['stellarator_parameters']
+num_phi = stellarator_parameters['num_phi']
+num_theta = stellarator_parameters['num_theta']
+
+# Prepare the build dictionary according to your requirements
+build = {
+    'phi_list': phi_list,
+    'theta_list': theta_list,
+    'wall_s': wall_s,
+    'radial_build': radial_build['components'],
+    'components': {}
+}
+
+# Convert thickness_matrix in 'radial_build' components to numpy arrays
+for component, details in radial_build['components'].items():
+    if 'thickness_matrix' in details and isinstance(details['thickness_matrix'], list):
+        build['components'][component] = {
+            'thickness_matrix': np.array(details['thickness_matrix'])
+        }
+        if 'h5m_tag' in details:
+            build['components'][component]['h5m_tag'] = details['h5m_tag']
+
+# Access VMEC file, magnets, source, and export configurations
+plas_eq = config['VMEC_file']
+magnets = config['magnet_coils']
+source = config['source_mesh_parameters']
+export = config['export_parameters']
+
+# Create stellarator using the configurations
+strengths = parastell.parastell(
+    plas_eq, build, repeat, num_phi, num_theta,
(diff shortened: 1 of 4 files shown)


example.py
```diff
@@ -0,0 +1,51 @@
...
+num_theta = stellarator_parameters['num_theta']
+
+# Prepare the build dictionary according to your requirements
+build = {
+    'phi_list': phi_list,
+    'theta_list': theta_list,
+    'wall_s': wall_s,
+    'radial_build': radial_build['components'],
+    'components': {}
+}
```
gonuke, line 29: Can't you just define this as `config['radial_build']` and then process the details into an NumPy array?

example.py
```diff
@@ -0,0 +1,51 @@
...
+
+# Convert thickness_matrix in 'radial_build' components to numpy arrays
+for component, details in radial_build['components'].items():
+    if 'thickness_matrix' in details and isinstance(details['thickness_matrix'], list):
```
gonuke, line 33: What if it's not a list? is in an error?

example.py
FusionSandwich: It seemed like it made more sense to have similar parameters grouped together in the yaml file. Even though they are apart of different inputs to parastell. I wasn't entirely sure if we'd be able to group the inputs for parastell together. And I didn't want to assign values to build that wouldn't be called elsewhere. But I can separate out variables from the radial_build in the yaml file like wall_s and repeated.

FusionSandwich: Then I can just set build as config['radial_build'] without the issues.

example.py
FusionSandwich: Yeah that was my experience, but I'll run some more tests and check it out again.

gonuke: This may be a question for @connoramoreno , but one wanted a uniform thickness blanket, how would it be defined in the current struture? Would it still be a theta/phi grid with very few points (2x2)? Or could it be a scalar for each thickness?  Do we have to handle cases like that?

example.py
gonuke: I think this may be worth a conversation as a group to balance the user experience vs the code benefits

example.py
FusionSandwich: I'll ask Conner about it before the Thursday meeting.

gonuke:
Please rebase this to capture updates to .gitignore.  I'd like to merge it ASAP

example.py
connoramoreno: Sorry, the notification for this got lost in my email inbox. In the current implementation, components of uniform thickness are still defined on a grid but all elements have equal value. We could choose to handle input for uniform thickness components differently, of course.

gonuke: We should probably issue an error if this data is NOT a list, then.

gonuke:
Something's fishy in this PR - there are too many files included that are not different from the target branch.  We may need to do a little git repair...

FusionSandwich:
Yeah haha I was having some weird issues rebasing it. Even Edgar was surprised with how it was acting.


________________________________

gonuke:
I don't know how it got to this state, but I was able to rebase it cleanly.  It would be great to sort this out today, even if you need the help of some other folks to navigate the git.

FusionSandwich:
Yeah I'm going to put this issue and the conversation about the parastell yaml file in the hacker within agenda.

________________________________

FusionSandwich:
Did we do what you expected with it?

lewisgross1296 (commented):
@FusionSandwich when you get a chance, like we discussed  "Yamlification of parastell" might also be a good name for this PR (Instead of "Initial Commit Yaml Example Script").

I think our rebase and force push likely addressed @gonuke's concern of extra files getting merged that this PR shouldn't have changed.

FusionSandwich:
Sounds good and thanks again for the help!
________________________________

gonuke (approved):
Thanks for cleaning up the PR @FusionSandwich

PR #58, opened 2024-03-26
Add dockerfile

### LINKED ISSUE #57: Create a docker image for testing that has a reasonable environment
To better support #12 we should confirm the best way to define a Dockerfile that makes a robust parastell testing environment.  The first version can skip the Cubit component, which makes it only really useful for source mesh testing.  We can then use that Dockerfile to build a CI image for testing in #13

Edgar-21:
Heavily inspired by a dockerfile shared by Shimwell here https://forum.coreform.com/t/trelis-in-docker/442/2.

Uses the miniconda docker image as the base (Debian Bullseye), found here https://hub.docker.com/r/continuumio/miniconda3. To use cubit, the image can be run interactively, and the license file copied to the appropriate location. Cubit works in terminal with the --no-graphics option and can be imported into python. 

I expect it will need a couple additions for this to work with Github actions. I'll start looking at that documentation in the next few days, in the meantime wanted to get this posted to get some eyes on it sooner rather than later

fixes #57

### DIFF
--- a/dockerfile
+++ b/dockerfile
@@ -1,7 +1,7 @@
 FROM continuumio/miniconda3
 
 ENV TZ=America/Chicago
-ENV PYTHONPATH=/opt/Coreform-Cubit-2024.3/bin/
+ENV PYTHONPATH=/opt/Coreform-Cubit-2023.11/bin/
 ENV PYTHONPATH=$PYTHONPATH:/opt/pystell_uw
 ENV PYTHONPATH=$PYTHONPATH:/opt/parastell
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
@@ -34,7 +34,7 @@ RUN apt-get install -y libgl1-mesa-glx \
                         libxinerama1
 
 # download cubit
-RUN wget -O /cubit.deb https://f002.backblazeb2.com/file/cubit-downloads/Coreform-Cubit/Releases/Linux/Coreform-Cubit-2024.3%2B46968-Lin64.deb
+RUN wget -O /cubit.deb https://f002.backblazeb2.com/file/cubit-downloads/Coreform-Cubit/Releases/Linux/Coreform-Cubit-2023.11%2B43088-Lin64.deb
 
 # install cubit
 RUN dpkg -i cubit.deb
--- a/dockerfile
+++ b/dockerfile
@@ -1,0 +1,42 @@
+FROM continuumio/miniconda3
+
+ENV TZ=America/Chicago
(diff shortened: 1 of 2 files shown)


Edgar-21:
perhaps this goes without saying but this will also need the parastell dependencies added, I was thinking about doing that with conda/mamba since that is how our install instructions are written

dockerfile
```diff
@@ -0,0 +1,42 @@
...
+                       curl \
+                       wget \
+                       libxm4 
+RUN apt-get clean
+
+# download cubit
+RUN wget -O /cubit.deb https://f002.backblazeb2.com/file/cubit-downloads/Coreform-Cubit/Releases/Linux/Coreform-Cubit-2024.3%2B46968-Lin64.deb
+
+
+# install dependencies
```
gonuke, line 26: Why not do these above?  Also, some are repeated from above

gonuke (changes requested):
Thanks @Edgar-21 - we might want to consolidate the dependencies

gonuke:
We need to add the environment file here too.

Edgar-21:
This has a conda environment named `parastell_env`. After adding the license and activating the environment, ExampleScript.py from the main branch will run to completion and generate the expected outputs, at least when I run it

environment.yml
```diff
@@ -0,0 +1,20 @@
...
+  - cadquery
+  - moab==5.5.0[build=nompi_tempest_*]
+  - cad_to_dagmc
+  - ca-certificates
+  - certifi
+  - openssl
+  - matplotlib
+  - pip:
+      - netcdf4
+      - pyyaml
```
gonuke, line 20: can we add `pytest` here for dev convenience?

dockerfile
```diff
@@ -0,0 +1,54 @@
...
+# download cubit
+RUN wget -O /cubit.deb https://f002.backblazeb2.com/file/cubit-downloads/Coreform-Cubit/Releases/Linux/Coreform-Cubit-2023.11%2B43088-Lin64.deb
+
+# install cubit
+RUN dpkg -i cubit.deb
+
+# parastell env
+COPY ./environment.yml /environment.yml
+
+RUN conda env create -f environment.yml
```
gonuke, line 45: I some suggestion that this might make `parastell_env` the default environment, but can't be sure.
```suggestion
RUN conda env create -f environment.yml

ENV CONDA_DEFAULT_ENV parastell_env

```

another alternative is to 
```
RUN echo "source activate parastell_env" >> ~/.bashrc
```

gonuke (changes requested):
I think this works for me.  I have a few convenience suggestions.

gonuke (approved):

gonuke:
Thanks @Edgar-21 - this works for me

PR #64, opened 2024-03-29
first attempt to rearrange files

### LINKED ISSUE #61: remove old code and confirm directory structure
Place new files in the right folders to be imported properly in both source code versions, installed versions and testing. (see also #62)

gonuke:
This is my first pass at getting the files in a more pythonic arrangement.

Fixed #61


parastell/tests/files_for_tests/coils.5Beta_2023_09_28_13_16_38
connoramoreno, line 1: This file and `coils.txt` are the same, so it is not necessary to have both. However, this does raise the question of which naming convention we'd prefer. It might be more better to use this one (or some different abbreviated version of it) since `coils.txt` isn't representative of coil optimization output; `coils.5Beta_2023_09_28_13_16_38` is the direct output of optimization and thus follows that naming convention.

I suggest we use something like `coils.5Beta` or `coils.example` to make it clearer to users what kind of file is needed. `.txt` might confuse users.

Examples/example_python.py
```diff
@@ -1,6 +1,6 @@
-import src.parastell as ps
 import numpy as np
 
+import parastell as ps
```
connoramoreno, line 3: As the file tree is set up, assuming the repository base directory is what's on the user's `PYTHONPATH`, the proper import syntax would be
```python
import parastell.parastell as ps
```
This would apply to the other example script as well, of course.

parastell/parastell.py
connoramoreno, line 1: Considering the parent directory is named `parastell`, we might consider renaming this file to avoid redundancy. As is (and mentioned above), importing this module in a python script uses the syntax
```python
import parastell.parastell
```
I suggest we rename it to something like `stellarator.py`

gonuke: As much as it looks redundant, I think this is a common way to do this

connoramoreno: If it's common then fair enough; I just didn't want to confuse users

connoramoreno:
Tests pass with this new arrangement (there was an issue with relative importing in the test scripts for the previous arrangement). A minor tweak is needed in `test_parastell.py` to use the proper coil file name (`coils.example` instead of `coils.txt`).

In the `Examples` directory, we'll want to rename the example coils file as before and modify the relevant lines in the example scripts. We'll also want to change how ParaStell is imported in the example scripts as mentioned above (`import parastell.parastell` instead of import `parastell`).

connoramoreno:
Thanks @gonuke, this looks good to me. Do we want to open this for review?

tests/test_parastell.py
connoramoreno, line 1: Sorry, looks like I missed that `coils.txt` needs to be changed to `coils.example` in the `magnets` dictionary here as well

connoramoreno, line 1: Sorry, looks like I missed that `coils.txt` needs to be changed to `coils.example` in the `magnets` dictionary here as well

connoramoreno (changes requested):
One more small change; sorry I missed this

connoramoreno (approved):
Passes all tests; looks good @gonuke

PR #65, opened 2024-03-29
Nwl area correction

### LINKED ISSUE #23: Implement area correction for theta-phi map
Current capability uses the same (unit?) surface area for each theta-phi patch.  This is not even true in idealized space - inboard patches are smaller than outboard patches - and even less true in real space.

Develop a way to approximate the area of each theta-phi patch, and implement it to calculate correct NWL values.

Edgar-21:
This adds area normalization to the NWL tool.

An array of bin boundaries in cartesian coordinates in generated. This array is then used to create an array of the area of each bin, calculated by creating 2 triangles of the 4 corners of each bin and summing their area. The energy of neutron crossings in each bin, which was previously calculated is then divided by the area in each bin.

Some other minor changes to import statements were made, as well as the addition of an argument that allows users to specify how many of the surface_source crossings they want to use for generating the plot. Defaults to using all crossings (added this because the surface source I was testing with took a couple hours to run, I left it in as it seemed potentially useful but certainly not opposed to taking it back out).

The total area of the first wall calculated was 233 m2, while in Cubit the first wall step file was measured to have an area of 229 m2. 

closes #23

### DIFF
--- a/NWL/NWL.py
+++ b/NWL/NWL.py
@@ -384,17 +384,46 @@ def plot(NWL_mat, phi_pts, theta_pts, num_levels):
 
     levels = np.linspace(np.min(NWL_mat), np.max(NWL_mat), num = num_levels)
     fig, ax = plt.subplots()
-    CF = ax.contourf(phi_pts, theta_pts, NWL_mat, levels = levels)
+    CF = ax.contourf(phi_pts, theta_pts, NWL_mat.T, levels = levels)
     cbar = plt.colorbar(CF)
     cbar.ax.set_ylabel('NWL (MW)')
     plt.xlabel('Toroidal Angle (degrees)')
     plt.ylabel('Poloidal Angle (degrees)')
     fig.savefig('NWL.png')
 
+def area_from_corners(corners):
+    """
+    Calculate an approximation of the area defined by 4 xyz points
+
+    Arguments:
+        corners (4x3 numpy array): list of 4 (x,y,z) points
+
+    Returns:
+        area (float): approximation of area
+    """
+    # triangle 1
+    v1 = corners[3] - corners[0]
+    v2 = corners[2] - corners[0]
+
+    v3 = np.cross(v1,v2)
+
+    area1 = np.sqrt(np.sum(np.square(v3)))/2
+
+    # triangle 2
+    v1 = corners[1] - corners[0]
+    v2 = corners[3] - corners[0]
+
+    v3 = np.cross(v1,v2)
+
+    area2 = np.sqrt(np.sum(np.square(v3)))/2
+
+    area = area1 + area2
+
+    return area
 
 def NWL_plot(
-    source_file, ss_file, plas_eq, tor_ext, pol_ext, wall_s, num_phi = 101,
-    num_theta = 101, num_levels = 10
+    source_file, ss_file, plas_eq, tor_ext, pol_ext, wall_s, num_phi = 100,
+    num_theta = 101, num_levels = 10, num_crossings = None
     ):
     """Computes and plots NWL.
 
@@ -408,6 +437,8 @@ def NWL_plot(
         num_phi (int): number of toroidal angle bins (defaults to 101).
         num_theta (int): number of poloidal angle bins (defaults to 101).
         num_levels (int): number of contour regions (defaults to 10).
+        num_particles (int): number of crossings to use from the surface source.
+            If none, then all crossings will be used
     """
     import read_vmec
     
@@ -415,6 +446,9 @@ def NWL_plot(
     pol_ext = np.deg2rad(pol_ext)
     
     coords = extract_coords(source_file)
+
+    if num_crossings is not None:
+        coords = coords[0:num_crossings]
     
     # Load plasma equilibrium data
     vmec = read_vmec.vmec_data(plas_eq)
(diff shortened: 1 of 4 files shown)


NWL/NWL.py
```diff
@@ -408,13 +438,18 @@ def NWL_plot(
         num_phi (int): number of toroidal angle bins (defaults to 101).
         num_theta (int): number of poloidal angle bins (defaults to 101).
         num_levels (int): number of contour regions (defaults to 10).
+        num_particles (int): number of crossings to use from the surface source.
```
gonuke, line 441: this doc doesn't match the parameter name

NWL/NWL.py
```diff
@@ -451,6 +486,41 @@ def NWL_plot(
...
+            
+            if theta < theta_min:
+                theta = theta_min
+            elif theta > theta_max:
+                theta = theta_max
+
+            if phi < phi_min:
+                phi = phi_min
+            elif phi > phi_max:
+                phi = phi_max
```
gonuke, line 505: First, why would this happen?  Doesn't the above code prevent this? If not, can't we prevent it?

Second, why not:
```
theta = min(max(theta_min, theta), theta_max)
phi = min(max(phi_min, phi), phi_max)
```

or even better? do this with numpy arrays outside of a loop:
```
theta_bins = min(max(theta_bins, theta_min), theta_max)
phi_bins = min(max(phi_bins, phi_min), phi_max)
```

NWL/NWL.py
```diff
@@ -451,6 +486,41 @@ def NWL_plot(
...
+
+            x,y,z = vmec.vmec2xyz(wall_s,theta,phi)
```
gonuke, line 507: I think the `source_mesh` code now has a way to generate a list of x,y,z for a list of theta, phi without a loop

NWL/NWL.py
```diff
@@ -384,17 +384,47 @@ def plot(NWL_mat, phi_pts, theta_pts, num_levels):
...
+
+    v3 = np.cross(v1,v2)
+
+    area1 = np.sqrt(np.sum(np.square(v3)))/2
+
+    # triangle 2
+    v1 = corners[1] - corners[0]
+    v2 = corners[3] - corners[0]
+
+    v3 = np.cross(v1,v2)
```
gonuke, line 417: This combination of corners is suspicious to me:
```
1 ---- 2
|      |
|      |
0 ---- 3

Current triangle 1:
      2
    /  |
  /    |
0 ---- 3

Current triangle 2:
 1
|   \ 
|     \
0 ---- 3

gonuke (changes requested):
A few little questions and one big one about the triangles.

I also feel as though more can be done with numpy functions/arrays instead of loops, but I'll have to try some things out

NWL/NWL.py
Edgar-21: You are correct, the 2d histogram should prevent this from occurring, thank you

NWL/NWL.py
Edgar-21: yup, that's my bad. Changing to have the two triangles as follows
0-----1 
|     /
3
and
       1
    /   |
 3----2

Edgar-21: scratch that, they are actually like this
 0
|   \ 
|     \
3 ---- 2 
and
0----1
 \     |
   \   |
      2

Edgar-21:
rebased onto oo_version

NWL/NWL.py
Edgar-21: I wasn't able to find anything like that, I saw a couple spots that used vmec2xyz but they were in loops. Maybe you can point me to the specific bit of code? Sorry if I missed it

gonuke: I guess it's in `invessel` and it's sort of a loop, but embedded in a function so the loop only needs to be written once, and it probably isn't worth relying on it here...

gonuke: That said, my gut says there is a way to do this all with numpy arrays that relies on fewer loops somehow...

gonuke:
Do we have any tests for any of this?

Edgar-21:
I can add a few tests, I think testing the full functionality will require adding support for NWL geometry to the oo_version and maybe there will be some issues with conflicts between parastell & openmc. I feel that's beginning to get out of the scope for what I initially set out to do with this PR (just add area calculation), maybe the scope needs to be revised or a second PR may be in order. 

I guess I see a few options, I can:
1. try to take on all of that here in this PR
2. add a couple minimal tests here, knowing that the current state does not integrate well with the oo_version. and we later take on individual issues related to integrating the NWL tool with the oo_version (adding support in parastell, retooling the NWL script, etc)
3. I close this PR and redo it but instead PR into main (where it will work) and go back to bullet 2 later (maybe I should have done this in the first place)

Option 2 or 3 would be my preference, and of those option 3 makes the most sense to me, open to whatever suggestions people may have outside of these as well

gonuke:
So does this version not work with the oo_version of parastell?

connoramoreno:
I agree that PRing into `main` might have been the more appropriate thing to do but I worry about what the resulting rebase of `oo_version` onto `main` might look like. `oo_version` is currently 68 commits ahead of `main`...

Edgar-21:
I'd start on a fresh branch of main if I was going to PR into it, not many changes need to be made. I had previously had some old files in my parastell directory that allowed the NWL script to work and didn't realize the problems I was going to have until after the re-arrangement and removal of old code

connoramoreno:
I guess this PR doesn't touch any of the "core" scripts, so a rebase might actually be seamless

Edgar-21:
would that move all the commit history of oo_version+commits made to this branch to main?

connoramoreno:
If you were to open a different PR where you make these changes to the NWL scripts in a branch based on `main` instead of `oo_version`, the commits would only apply to `main`. Then, since `oo_version` is also based on `main`, `oo_version` would need to be rebased in order to incorporate the commits changing the NWL scripts.

connoramoreno:
But let's wait for Paul's input before potentially creating a different PR

gonuke:
So looking at the NWL stuff again, I see that we passed it all over in the `oo_version` update.  This PR probably makes most sense against `main` because I think (?) that's the version of parastell it needs to work, fundamentally.

We'll want a different PR to update NWL for the new parastell model, possibly with some simplifications?

Edgar-21:
closing PR, will open a new one against main

PR #70, opened 2024-03-30
YAML reader function

### LINKED ISSUE #52: Review structure of YAML file
Consider improvements in yaml file structure to reduce redundancy and/or opportunities for inconsistency

### LINKED ISSUE #54: Standardize code patterns for reading input files

connoramoreno:
Changes the location of the YAML reader function `read_yaml_config`. Previously, each ParaStell script had an identical `read_yaml_config` function to read YAML input. This PR moves this function to `utils.py`, and each script then imports it.

Closes #52 
Closes #54


connoramoreno:
It should be noted that this is a draft since the relative imports in each script seem to be causing issues when using YAML input from the command line. For example, assuming parastell.py and config.yaml are in the same directory, the command
```
python3 parastell.py config.yaml
```
results in the following error:
```
Traceback (most recent call last):
  File "/home/camoreno/parastell/parastell/parastell.py", line 6, in <module>
    from . import log
ImportError: attempted relative import with no known parent package
```

gonuke:
I think we'll want to get to a point where we actually install this rather than just run from the source directory.  Alternatively, we put the `parastell` directory in the python path?

connoramoreno:
This is with `parastell` on `PYTHONPATH`, specifically the repository directory. This also uses the updated file structure.

connoramoreno:
Executing from a different directory with `config.yaml`,
```
python3 ~/parastell/parastell/parastell.py config.yaml
```
results in the same error.

gonuke:
If that fixes it, I guess let's change it. Separately, I'd like to better understand these relative imports, because they should work.

connoramoreno:
They work when ParaStell is imported as a module, but not when run directly with a YAML input.

connoramoreno:
It might be that relative imports are defined only for Python modules, but when these scripts are run from command line with a YAML argument, they might not be considered modules by Python anymore

gonuke:
Yes... so I think the right way to do this is to generate a lightweight script that imports `parastell` and calls `parastell.parastell()`

if you make a file called `stellarator` with these contents and `chmod u+x`, it works.
```
#!/usr/bin/env python3

from parastell import parastell

parastell.parastell()
```

and make it executable.

I think there are simple ways to do this with setup tools as part of installation. But that may be for another time.

connoramoreno:
I am planning to lump in the solution we discussed for #52 into this PR since it's so related, unless there are any objections.

connoramoreno:
Also considering rebasing this onto `null-logger` since there are some changes in that branch that affect the YAML input workflow.

executables/invessel_build.py
gonuke, line 1: Maybe name these without the `.py` extension?

gonuke, line 1: Also, if we figure out how to use `pip install` correctly, these can be automatically generated in the right place in the path.

gonuke (changes requested):
Just a minor comment on naming the executables.

executables/invessel_build.py
connoramoreno: Would we rather make ParaStell pip-installable or condo-installable, given the dependency on conda packages? I'm not sure that some of our dependencies can be installed via pip.

gonuke: `pip` can also be used to install the software from the source code in a local place on the Python path.   This is the mode I'm thinking of, not necessarily installing with `pip` from PyPi

gonuke (approved):
LGTM - thanks @connoramoreno

PR #74, opened 2024-04-04
Update installation instructions on `main`

connoramoreno:
Updates details about dependencies in the installation instructions on `main`.

The installation instructions did not accurately reflect ParaStell's Python dependencies. Additional Python dependencies have been added.

Furthermore, this changes installation instructions for PyStell-UW to specify cloning the `old` branch. There are several issues with the `master` branch of PyStell-UW that prevent its successful integration with ParaStell.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,86 +1,80 @@
-## ParaStell
-Parametric 3D CAD modeling tool for stellarator fusion devices. This toolset takes VMEC plasma equilibrium data to build intra-magnet components of varying thickness from a user-specified radial build, as well as corresponding coil filament point-locus data to build magnet coils of user-specified cross-section.
+# ParaStell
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,86 +1,80 @@
...
+### Install PyStell-UW
+Download and extract the PyStell-UW repository:
 
-```
-git clone https://github.com/aaroncbader/pystell_uw.git
+```bash
+git clone --single-branch --branch old https://github.com/aaroncbader/pystell_uw.git
 ```
 
-or download the and extract the zip from [pystell_uw](https://github.com/aaroncbader/pystell_uw). Once extracted, add the directory to the `PYTHONPATH`.
```
Edgar-21, line 75: perhaps update this link to go to the old branch, and maybe add a note about why the old branch is necessary, though I would expect most folks will use git to clone rather than downloading the zip
https://github.com/aaroncbader/pystell_uw/tree/old

Edgar-21 (changes requested):
This looks good to me otherwise, have you tested these install instructions? Looks the same as whats in the environment.yml file so I would imagine it works the same

Edgar-21 (approved):
lgtm thanks

PR #76, opened 2024-04-04
added area normalization to NWL.py

### LINKED ISSUE #23: Implement area correction for theta-phi map

Edgar-21:
Calculates the area of each bin by converting the corners to xyz and approximating the area as 2 triangles.

The following plot was generated using NWL_plot.py, and a surface source with a million particles.
image

Importing the points in `bin_arr` to cubit and overlaying on the geometry it can be seen that they align nicely with the geometry, with half size bins at the toroidal and poloidal extent edges
image

The area from summing `area_array` for the parameters in NWL_plot.py is 228.97m2, and measuring in cubit the area is 229.00 m2.

Fixes #23

### DIFF
--- a/NWL/NWL.py
+++ b/NWL/NWL.py
@@ -384,7 +384,7 @@ def plot(NWL_mat, phi_pts, theta_pts, num_levels):
 
     levels = np.linspace(np.min(NWL_mat), np.max(NWL_mat), num = num_levels)
     fig, ax = plt.subplots()
-    CF = ax.contourf(phi_pts, theta_pts, NWL_mat, levels = levels)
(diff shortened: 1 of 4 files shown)


connoramoreno:
Beautiful. The plot features the area normalization right? It's missing from the NWL units.

Edgar-21:
yes, that's my bad I did not look closely enough at the plotting function

gonuke (approved):
Thanks @Edgar-21

Issue #79, opened 2024-04-09
Edge Recovery Error in CAD-to-DAGMC Workflow


connoramoreno:
When trying to export a DAGMC model of the test in-vessel build geometry using the `export_cad_to_dagmc` method of the `InVesselBuild` class, I am running into the following error during the 2-D meshing step:
```bash
Info    : Meshing 2D...
Info    : [  0%] Meshing surface 1 (BSpline surface, MeshAdapt)
Info    : [ 10%] Meshing surface 2 (Plane, MeshAdapt)
Info    : [ 20%] Meshing surface 3 (Plane, MeshAdapt)
Info    : [ 30%] Meshing surface 4 (BSpline surface, MeshAdapt)
Info    : [ 40%] Meshing surface 5 (Plane, MeshAdapt)
Info    : [ 50%] Meshing surface 6 (Plane, MeshAdapt)
Info    : [ 60%] Meshing surface 7 (BSpline surface, MeshAdapt)
Info    : [ 70%] Meshing surface 8 (Plane, MeshAdapt)
Info    : [ 80%] Meshing surface 9 (Plane, MeshAdapt)
Info    : [ 90%] Meshing surface 10 (BSpline surface, MeshAdapt)
Warning : Impossible to recover edge 1496 1496 (error tag -1)
Warning : Impossible to recover edge 4 4 (error tag -1)
Error   : Unable to recover the edge 1500 (364/364) on curve 4 (on surface 10)
```
I am unsure if this error is specific to ParaStell or CAD-to-DAGMC. The geometry used is that of the test in-vessel build; that is, the geometry includes a plasma (yellow), scrape-off layer (green), and a single generic component (purple/pink), pictured below.

<img width="240" alt="Screenshot 2024-04-08 at 6 59 09 PM" src="https://github.com/svalinn/parastell/assets/67344181/dc0029f8-9626-44a6-8150-c5164f47e742"><img width="240" alt="Screenshot 2024-04-08 at 6 59 39 PM" src="https://github.com/svalinn/parastell/assets/67344181/9b90d5ff-7095-444b-bed4-9a19840ee104">

[...]

PR #80, opened 2024-04-09
Update integration with CAD-to-DAGMC v0.6.2 for `oo_version`

### LINKED ISSUE #78: Integration with Updated CAD-to-DAGMC
CAD-to-DAGMC had a recent update that slightly changes its syntax. ParaStell's `main` branch was updated to reflect this change thanks to PR #77, but the `oo_version` branch will need the same changes.

connoramoreno:
Updates syntax in `export_cad_to_dagmc` method of `InVesselBuild` class for the `oo_version` branch to reflect update to CAD-to-DAGMC dependency. The test script for `InVesselBuild` is also updated to incorporate testing of the in-vessel build DAGMC export.

It should be noted that, at the initial commit for this PR, the test does not pass due to potential geometry issues. This issue is documented in #79. The new syntax incorporated is correct but we may want to wait until #79 is resolved before considering this PR for review.

Closes #78.


connoramoreno:
With the changes to unit tests introduced in PR #81, the issue documented in #79 no longer occurs. Are we comfortable considering the issue resolved, at least as far as this PR is concerned?

gonuke:
I think that's probably OK, but maybe we don't close #79 yet...

connoramoreno:
That makes sense to me!

parastell/invessel_build.py
```diff
@@ -352,15 +352,18 @@ def export_cad_to_dagmc(self, filename='dagmc', export_dir=''):
...
-        for name, component in self.Components.items():
-            model.add_cadquery_object(
-                component, material_tags=[self.radial_build[name]['mat_tag]']]
-            )
+        for component in self.Components.values():
+            model.add_cadquery_object(component)
+        
+        material_tags = [
+            component['mat_tag'] for component in self.radial_build.values()
+        ]
```
gonuke, line 360: Do the relative orders of these things matter?  One loops over `self.Components` and one over `self.radial_build`.  Presumably those things are both built in the same order, but nothing guarantees it...?

gonuke (changes requested):
One question/concern here.... not sure if it exists in the PR we already merged for the old  version.  It seems like the decision of `cad_to_dagmc` bind this data together more loosely is dangerous and risks material tags being out of sync with the objects. (@shimwell ?)

gonuke:
It's not just about the number, it's about the order!  This API relies on the user to keep track of this data in a way that preserves this binding.

PR #81, opened 2024-04-09
Improve Unit Testing

### LINKED ISSUE #73: Streamline unit tests
We have added units tests with the migration to object-oriented code, but they have become rather cumbersome and slow running.  We should evaluate each of these tests and see how we can make them faster.

connoramoreno:
Simplifies the geometry generated for the ParaStell, in-vessel build, and magnet tests and incorporates pytest fixtures for all tests.

All tests pass using standard pytest command-line functionality. The full test suite takes 4 minutes, 47 seconds to complete.

Closes #73.


gonuke (approved):
Thanks @connoramoreno - this is a big speed up!

PR #84, opened 2024-04-10
Add Github action for building a Docker image for testing

gonuke:
This adds capabilities to use Github actions to build a docker image that we can then use for CI unit testing.

* builds the docker image only including dependencies
* uses that docker image to run tests
    * checks out the repo
    * saves Github secretes in the environment for the RLM server and RLM password
    * uses `sed` to update the license file with those secrets
    * runs pytests

### DIFF
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -55,7 +55,7 @@ jobs:
 
       - name: test secrets
         run: |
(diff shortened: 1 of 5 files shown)


gonuke:
Making progress here - two of the tests work fine.

gonuke:
Withdrawing this because @connoramoreno replicated it in #95

PR #85, opened 2024-04-11
Update Inputs for User-Facing Classes

connoramoreno:
Reduces arguments of all user-facing classes to a minimal list and introduces setters and getters for those classes. Additionally, implements the use of `**kwargs` to handle optional setting of default class attributes for both the Python API and command-line interface with YAML input.

Furthermore, introduces a `RadialBuild` class in `invessel_build.py` to minimize the argument list for `InVesselBuild`.


connoramoreno:
I have another commit coming soon that will improve the use of `kwargs`, allowing only certain `kwargs` to be supplied

connoramoreno:
PR is ready for review now!

parastell/invessel_build.py
```diff
@@ -49,134 +53,99 @@ def orient_spline_surfaces(volume_id):
...
+    @repeat.setter
+    def repeat(self, num):
+        self._repeat = num
+        if (self._repeat + 1) * self.radial_build.toroidal_angles[-1] > 360.0:
             e = AssertionError(
                 'Total toroidal extent requested with repeated geometry '
-                'exceeds 360 degrees. Please examine phi_list and the repeat '
-                'parameter.'
+                'exceeds 360 degrees. Please examine the \'repeat\' parameter '
+                'and the \'toroidal_angles\' parameter of \'radial_build\'.'
```
gonuke, line 144: You can avoid having to escape these single quotes by using the other quote mark from that defining the string
```suggestion
                'exceeds 360 degrees. Please examine the "repeat" parameter '
                'and the "toroidal_angles" parameter of "radial_build".'
```

parastell/utils.py
```diff
@@ -56,124 +58,63 @@ def expand_ang_list(ang_list, num_ang):
...
+            'all_kwargs' is True, a logger object should be supplied.
+
+    Returns:
+        kwarg_dict (dict): dictionary of keyword arguments and values.
+    """
+    kwarg_dict = {}
+    for name, value in dict.items():
+        if name in allowed_kwargs:
+            kwarg_dict.update({name: value})
+        elif all_kwargs and name not in allowed_kwargs:
```
gonuke, line 88: Isn't the second half of this condition implied by the fact that you didn't match on the `if`?

gonuke (changes requested):
This is a bigger upgrade than expected - thanks for all the effort @connoramoreno

parastell/invessel_build.py
```diff
@@ -49,134 +50,94 @@ def orient_spline_surfaces(volume_id):
...
+        self.vmec_obj = vmec_obj
+        self.radial_build = radial_build
+        
+        self.repeat = 0
+        self.num_ribs = 61
+        self.num_rib_pts = 67
+        self.scale = m2cm
+
+        for name, value in kwargs.items():
+            self.__setattr__(name, value)
```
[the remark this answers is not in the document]
gonuke: I think you took care of this already...

Examples/example_python.py
```diff
@@ -2,105 +2,109 @@
...
-    'num_theta': 81,
-    'num_phi': 61,
-    'toroidal_extent': 90.0,
-    'scale': 100,
-    'filename': 'source_mesh',
-    'export_dir': export_dir
-}
+num_s = 4
+num_theta = 8
+num_phi = 4
```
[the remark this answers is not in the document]
connoramoreno: I think with sufficient documentation, a tuple input would be fine

parastell/utils.py
connoramoreno: Well I want the first condition to happen whether `all_kwargs` is `True` or `False`, while I want the error to be raised only under the condition that `all_kwargs = True`.

This functionality is particularly useful when extracting keyword arguments from YAML-derived dictionaries, since not all keys in those dictionaries are keyword arguments. In such cases, we set `all_kwargs = False` since we don't want an error to be raised when the algorithm comes across a key that is a positional argument.

gonuke: But I think you only get to this check against `all_kwargs` if `name not in allowed_kwargs`... right?
```suggestion
        elif all_kwargs:
```

connoramoreno: Oh, I see what you're saying. Good point! I thought the "second half of the condition" you were referring to was the `elif` as a whole.

parastell/invessel_build.py
```diff
@@ -536,6 +517,221 @@ def generate_rib(self):
...
+        self._wall_s = s
+        if self._wall_s < 1.0:
+            e = ValueError(
+                'wall_s must be greater than or equal to 1.0.'
+            )
+            self._logger.error(e.args[0])
+            raise e
+        
+        if hasattr(self, 'radial_build'):
+            self.radial_build = self.radial_build
```
gonuke, line 642: What is this doing?

parastell/invessel_build.py
```diff
@@ -536,6 +517,221 @@ def generate_rib(self):
...
+    @property
+    def radial_build(self):
+        return self._radial_build
+    
+    @radial_build.setter
+    def radial_build(self, build_dict):
+        self._radial_build = build_dict
+        
+        if self._wall_s == 1.0 and 'sol' in self._radial_build:
+            del self.radial_build['sol']
```
gonuke, line 653: Doesn't this prevent someone from defining a SOL with varying thicknesses?

parastell/invessel_build.py
```diff
@@ -536,6 +517,221 @@ def generate_rib(self):
...
+    @sol_mat_tag.setter
+    def sol_mat_tag(self, mat_tag):
+        self._sol_mat_tag = mat_tag
+        self._set_mat_tag('sol', self._sol_mat_tag)
+
+    def _set_mat_tag(self, name, mat_tag):
+        """Sets material tag for a given component.
+        (Internal function not intended to be called externally)
+        """
+        self._radial_build[name]['mat_tag'] = mat_tag
```
gonuke, line 734: Should this check to see if the component `name` has a `thickness` matrix already?  Maybe it doesn't matter?  It only matters that the `thickness` matrix is there once you aim to use it...

parastell/utils.py
```diff
@@ -56,124 +58,63 @@ def expand_ang_list(ang_list, num_ang):
...
+    """Constructs a dictionary of keyword arguments with corresponding values
+    for a class method from an input dictionary based on a list of allowable
+    keyword argument argument names. Conditionally raises an exception if the
+    user flags that the supplied list of allowable arguments should represent
+    all keys present in the input dictionary of arguments.
+
+    Arguments:
+        dict (dict): dictionary of arguments and corresponding values.
+        allowed_kwargs (list of str): list of allowed keyword argument names.
+        all_kwargs (bool): flag to indicate whether 'allowed_kwargs' should
```
gonuke, line 74: Should this default to `False`?

parastell/utils.py
```diff
@@ -56,124 +58,63 @@ def expand_ang_list(ang_list, num_ang):
...
+                f'"{fn_name}"'
+            )
+            logger.error(e.args[0])
+            raise e
+
+    return kwarg_dict
+
+
+def set_kwarg_attrs(
+    class_obj, kwargs, allowed_kwargs, logger
```
gonuke, line 100: is it reasonable to assume that the `logger` should always be `class_obj.logger` and not need to pass in an extra variable?

parastell/parastell.py
```diff
@@ -47,251 +48,315 @@ class Stellarator(object):
...
+        self._vmec_file = file
+        self._vmec_obj = read_vmec.VMECData(self._vmec_file)
+        if Path(self._vmec_file).suffix != '.nc':
+            e = ValueError(
+                'Plasma equilibrium VMEC data file input has extension '
+                f'"{Path(self._vmec_file).suffix}". File format must be '
+                'netCDF (".nc").'
+            )
+            self._logger.error(e.args[0])
+            raise e
```
gonuke, line 83: This could be limiting... I'm not sure we want to be this heavy-handed with file extensions.

gonuke: Won't `read_vmec` complain for this anyway, if the file is the wrong format?

parastell/invessel_build.py
```diff
@@ -49,134 +53,99 @@ def orient_spline_surfaces(volume_id):
...
+
+    @property
+    def logger(self):
+        return self._logger
+    
+    @logger.setter
+    def logger(self, logger_object):
+        self._logger = logger_object
+        if self._logger == None or not self._logger.hasHandlers():
+            self._logger = log.init()
```
gonuke, line 131: Since we are doing this in every class, what if we add `log.check_init(logger_object)`
```
# in log.py
def check_init(logger_object):
    if logger_object != None and not logger_object.hasHandlers():
        return logger_object
    else
       return log.init()
```

And this method then is just one line:
```
self._logger = log.check_init(logger_object)
```

gonuke (changes requested):
Some final comments as we close in on the final version.

parastell/invessel_build.py
connoramoreno: It's redefining the `radial_build` dictionary to reflect any changes in `wall_s`

parastell/invessel_build.py
connoramoreno: Yes, you're right. I'll need to come up with some way to delete `'sol'` from `radial_build` only under the right conditions

parastell/invessel_build.py
connoramoreno: The setter for `radial_build` checks that each component has a `'thickness_matrix'` of the proper shape, which indirectly checks that a `'thickness_matrix'` exists.

parastell/utils.py
connoramoreno: I'm not opposed to setting a default for it; would be one less argument to add in some of the function calls, and defaulting to `False` seems appropriate.

parastell/invessel_build.py
gonuke: Or maybe non-existent?

parastell/parastell.py
connoramoreno: It doesn't seem like `read_vmec` has an exception built-in for this, but I imagine netCDF4 might. The file gets read as
```python
from netCDF4 import Dataset
...
self.data = Dataset(fname)
```
where `fname` is the VMEC file name.

gonuke (approved):
This is ready to go! Great effort @connoramoreno!

PR #88, opened 2024-04-16
Introduce Null Logger Class

### LINKED ISSUE #46: Review approach for logger
The approach for using a logger across these methods should be reviewed to ensure it is sufficiently flexible for various use cases including:
* running as stand alone scripts
* new scripts that import all the different components
* logger-free use (?)

connoramoreno:
Adds a pseudo logger class, `NullLogger`, with some of the same class methods as the built-in Python `logging` class object, but whose methods do nothing when called.

The purpose of `NullLogger` is to streamline ParaStell code but not enforce logging when using the Python API. A logger does get enforced upon the user when using the command-line interface with YAML input.

Depends on #85.

Closes #46.

### DIFF
--- a/parastell/invessel_build.py
+++ b/parastell/invessel_build.py
@@ -742,6 +742,15 @@ def parse_args():
         'filename',
         help='YAML file defining ParaStell in-vessel component configuration'
     )
+    parser.add_argument(
+        '-l', '--logger',
+        default=False,
+        help=(
+            'Flag to indicate whether to instantiate a logger object (default: '
+            'False)'
+        ),
+        metavar=''
+    )
 
     return parser.parse_args()
 
@@ -763,7 +772,10 @@ def generate_invessel_build():
 
     vmec_file, invessel_build_dict = read_yaml_config(args.filename)
 
-    logger = log.check_init(None, null_logger=False)
+    if args.logger == True:
+        logger = log.init()
+    else:
+        logger = log.NullLogger()
 
     vmec_obj = read_vmec.VMECData(vmec_file)
 
--- a/parastell/invessel_build.py
+++ b/parastell/invessel_build.py
@@ -671,6 +671,8 @@ class RadialBuild(object):
         }
         
         for name, component in self._radial_build.items():
+            component['thickness_matrix'] = \
+                np.array(component['thickness_matrix'])
             if (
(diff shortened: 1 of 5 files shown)


gonuke:
Let's try to finish #85 first...

parastell/invessel_build.py
```diff
@@ -761,6 +763,8 @@ def generate_invessel_build():
 
     vmec_file, invessel_build_dict = read_yaml_config(args.filename)
 
+    logger = log.check_init(None, null_logger=False)
```
gonuke, line 766: I'm not sure we want to use `check_init` here.  This is our script and we can actively decide whether we want a logger or not, perhaps based on a parser argument.

parastell/magnet_coils.py
```diff
@@ -609,6 +609,8 @@ def generate_magnet_set():
 
     magnet_coils_dict = read_yaml_config(args.filename)
 
+    logger = log.check_init(None, null_logger=False)
```
gonuke, line 612: Similar comment to invessel build - let's actively make a logger if we want one rather than rely on this default approach

parastell/parastell.py
```diff
@@ -494,7 +494,12 @@ def parastell():
         vmec_file, invessel_build, magnet_coils, source_mesh, dagmc_export
     ) = read_yaml_config(args.filename)
 
-    stellarator = Stellarator(vmec_file)
+    logger = log.check_init(None, null_logger=False)
```
gonuke, line 497: Ibid: actively make a logger in a script we own/control

parastell/source_mesh.py
```diff
@@ -430,6 +430,8 @@ def generate_source_mesh():
 
     vmec_file, source_mesh_dict = read_yaml_src(args.filename)
```
gonuke, line 432: Actively choose a logger or not

gonuke (changes requested):
I think we should only use `check_init` in our class initializers

parastell/invessel_build.py
gonuke: Specifically, we can call `log.init()` since we know we want a logger instead of `log.check_init()` with arguments that force a logger to be created.

gonuke (approved):
Thanks @connoramoreno

PR #91, opened 2024-04-18
return parameters used to plot NWL from NWL_plot

### LINKED ISSUE #90: NWL_plot should return the NWL matrix
It takes a long time to generate the NWL matrix and it would be a shame to have to do it all over again

Edgar-21:
The full function may take a long time to run, and it would be convenient to be able to avoid needing to rerun the entire function in order to iterate plots.

Closes #90 

Perhaps it would be better to split the NWL_plot into two functions, one that makes the NWL matrix and one that plots it?


gonuke (approved):
Nice quick addition @Edgar-21

PR #96, opened 2024-04-26
add normal test for CI

gonuke:


gonuke:
This should be the last thing before transitioning to the OO version

connoramoreno:
I'm still trying to wrap my mind around using GitHub Actions for CI. What's the purpose of having tests triggered by both pull requests and commit pushes? And something I'm still slightly unclear about is which version of the code is being tested for each GitHub Action workflow.

gonuke:
The other test only runs when a push happens and one of those two specific files changes: the GH action itself or the Dockerfile.

This test only runs on a pull request when any file changes.  We could, technically, NOT run this test when those two files change.

connoramoreno (approved):
Looks good to me! Thanks @gonuke

PR #98, opened 2024-04-27
Streamline kwargs

gonuke:
As I started looking at implementing NWL in the OO version, I realized that we were kind of heavy-handed with kwargs. I think this finds a better balance. Let me know what you think.


gonuke:
Closing in favor of an upstream branch rather than fork

PR #99, opened 2024-04-28
Streamline kwargs

gonuke:
As I started looking at implementing NWL in the OO version, I realized that we were kind of heavy-handed with kwargs. I think this finds a better balance. Let me know what you think.

Some explanations for the changes:
1. The old code had the lists of allowable arguments defined in multiple places.  They should be defined in as few places as possible - ideally only one.
2. Most of the uses of `construct_kwargs_from_dict` were for passing into a constructor for some class.  That class is the best place to know what arguments are allowable, so its best to pass in the whole dictionary and then cherry-pick relevant data in the constructor.
3. `set_kwarg_attrs` was the function that cherry-picked from the dictionary, but raised exceptions by default for extra arguments.  We probably discussed this at the time, and I may have recommended or endorsed it, but upon reflection it seems heavy-handed.  It is only because of this that we had to carefully curate the `kwargs` that we passed in the first place and maybe its fine to just silently ignore extra `kwargs`.
4. Once `set_kwarg_attrs` became simpler, I decided to just put the code directly into the constructors rather than as a function.  (this is perhaps dubious in our normal spirit of reuse...)
5. There are a few places that still need to pass in values extracted from dictionaries into function argument lists. Here it is more important to extract them more carefully, but this can be done inline.  I changed the method name from `construct_kwargs_from_dict` to `filter_kwargs` as this is more what it does and makes the name shorter.

### DIFF
--- a/parastell/parastell.py
+++ b/parastell/parastell.py
@@ -412,77 +412,6 @@ class Stellarator(object):
                 **kwargs
             )
 
-    def build_full_model(self, invessel_build, magnet_coils, source_mesh):
-        """Use all of the input to make a complete stellarator model with
-        in-vessel components, magnets, and a source mesh.
-
-        Arguments:
-            invessel_build (dict): dictionary of RadialBuild and InVesselBuild
-                parameters.
-            magnet_coils (dict): dictionary of MagnetSet parameters.
-            source_mesh (dict): dictionary of SourceMesh parameters.
-        """
-        self.construct_invessel_build(
-            invessel_build['toroidal_angles'],
-            invessel_build['poloidal_angles'],
-            invessel_build['wall_s'],
-            invessel_build['radial_build'],
-            **invessel_build
-        )
-
-        self.construct_magnets(
-            magnet_coils['coils_file'],
-            magnet_coils['cross_section'],
-            magnet_coils['toroidal_extent'],
-            **magnet_coils
-        )
-
-        self.construct_source_mesh(
-            source_mesh['mesh_size'],
-            source_mesh['toroidal_extent'],
-            **source_mesh
-        )
-
-    def export_full_model(
-        self, invessel_build, magnet_coils, source_mesh, dagmc_export,
-        export_dir=''
-    ):
-        """Exports a complete stellarator model when in-vessel components,
-        magnets, and source mesh have been constructed.
-
-        Arguments:
-            invessel_build (dict): dictionary of RadialBuild and InVesselBuild
-                parameters.
-            magnet_coils (dict): dictionary of MagnetSet parameters.
-            source_mesh (dict): dictionary of SourceMesh parameters.
-            dagmc_export (dict): dictionary of DAGMC export parameters.
-            export_dir (str): directory to which output files are exported
-                (optional, defaults to empty string).
-        """
-        export_cad_to_dagmc = invessel_build.get('export_cad_to_dagmc', False)
-        dagmc_filename = invessel_build.get('dagmc_filename', 'dagmc')
-
-        self.export_invessel_build(
-            export_dir=export_dir,
-            export_cad_to_dagmc=export_cad_to_dagmc,
-            dagmc_filename=dagmc_filename
-        )
-        
-        self.export_magnets(
-            export_dir=export_dir,
-            **(filter_kwargs(magnet_coils, mc.export_allowed_kwargs))
-        )
-
-        self.export_source_mesh(
-            export_dir=export_dir,
-            **(filter_kwargs(source_mesh, sm.export_allowed_kwargs))
-        )
-        
-        self.export_dagmc(
-            export_dir=export_dir,
-            **dagmc_export
-        )
-
 
 def parse_args():
     """Parser for running as a script.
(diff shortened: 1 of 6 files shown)


parastell/parastell.py
```diff
@@ -457,6 +412,63 @@ def export_dagmc(
...
+        self.export_invessel_build(
+            export_dir=export_dir,
+            export_cad_to_dagmc=export_cad_to_dagmc,
+            dagmc_filename=dagmc_filename
+        )
+        
+        self.export_magnets(
+            export_dir=export_dir,
+            **(filter_kwargs(magnet_coils,mc.export_allowed_kwargs))
+        )
```
connoramoreno, line 459: Why do we handle `export_invessel_build` and `export_magnets` differently? They handle keyword arguments very similarly.

parastell/parastell.py
```diff
@@ -605,107 +617,16 @@ def parastell():
...
-    ]
-    ivb_construct_kwargs = construct_kwargs_from_dict(
-        invessel_build,
-        ivb_construct_allowed_kwargs
-    )
+    stellarator.export_full_model(args.export_dir,
+                                  all_data['invesel_build'],
+                                  all_data['magnet_coils'],
+                                  all_data['source_mesh'],
+                                  all_data['dagmc_export'])
```
connoramoreno, line 628: Nice. Much shorter.

parastell/parastell.py
```diff
@@ -457,6 +412,63 @@ def export_dagmc(
...
+
+        # Source Mesh
+        self.construct_source_mesh(
+            source_mesh['mesh_size'],
+            source_mesh['toroidal_extent'],
+            **source_mesh
+        )
+
+    def export_full_model(self, export_dir, invessel_build, 
+                          magnet_coils, source_mesh, dagmc_export):
```
connoramoreno, line 445: We should probably include a docstring here. I'm happy to do so myself, if you prefer; I just need to figure out pushing commits to others' PRs...

parastell/parastell.py
```diff
@@ -457,6 +412,63 @@ def export_dagmc(
                 **kwargs
             )
 
+    def build_full_model(self, invessel_build, magnet_coils, source_mesh):
+        """Use all of the input to make a complete stellarator model
+        with invessel components, magnets, and a source mesh
+        """
```
connoramoreno, line 418: Should probably include some version of argument explanations in this docstring.

connoramoreno (changes requested):
These code changes are pretty slick; thanks for streamlining things @gonuke. I just have a question about exports in `parastell.py` and some requests regarding docstrings. I'll try to push a commit to this PR for the docstrings.

connoramoreno:
In my commit, I changed `export_dir` in `export_full_model` to a keyword argument to match our export method convention, in case Python users would like to use `build_full_model` and/or `export_full_model`.

However, using these methods requires inputs to be defined in dictionary format, which is different from how we've defined inputs for the rest of the Python API. It might be best just to enforce one canonical input format for Python users, and maintain that `build_full_model` and `export_full_model` be intended for the command-line workflow.

gonuke:
> However, using these methods requires inputs to be defined in dictionary format, which is different from how we've defined inputs for the rest of the Python API. It might be best just to enforce one canonical input format for Python users, and maintain that `build_full_model` and `export_full_model` be intended for the command-line workflow.

Right.  My first pass just made these global functions and took `stellarator` as an argument rather than member functions.  We could revert to that - what do you think?  Note: as I work towards the NWL mode, these two functions get moved to a method that is just for the main mode.  (see #100 as a work in progress)

connoramoreno:
I think making `build_full_model` and `export_full_model` global functions makes sense. We can then do something similar for the NWL PR. I can make that change to this PR quickly.

parastell/parastell.py
```diff
@@ -574,6 +529,63 @@ def check_inputs(
...
+    Arguments:
+        stellarator (object): Stellarator class object.
+        invessel_build (dict): dictionary of RadialBuild and InVesselBuild
+            parameters.
+        magnet_coils (dict): dictionary of MagnetSet parameters.
+        source_mesh (dict): dictionary of SourceMesh parameters.
+    """
+    stellarator.construct_invessel_build(**invessel_build)
+    stellarator.construct_magnets(**magnet_coils)
+    stellarator.construct_source_mesh(**source_mesh)
```
gonuke, line 547: @connoramoreno  These all have positional arguments - shouldn't we call them with those arguments extracted from the dictionaries explicitly here?

connoramoreno: Defining the positional arguments and unpacking the respective dictionaries in the same function call defines those positional arguments twice, causing a `TypeError`

connoramoreno: Just unpacking the dictionaries will appropriately set the positional arguments.

connoramoreno: If we want to explicitly set positional arguments, we'll need to do something similar to what I had before where secondary dictionaries are created for the `kwargs`

gonuke: OK - Is there an error if one of the positional arguments is missing?  (Maybe we should add a test for that at some point)

connoramoreno: E.g., I removed `poloidal_angles` from `config.yaml` to test:
```
TypeError: Stellarator.construct_invessel_build() missing 1 required positional argument: 'poloidal_angles'
```

gonuke: I think this is ready to merge, then...

connoramoreno: Agreed!

connoramoreno (approved):
Looks good now.

PR #106, opened 2024-05-17
Custom source prototype

gonuke:
Here's my prototype for allowing a custom source function. I haven't tested it and may be abusing Python language features so it may need some work to bring it to fruition!


connoramoreno:
Can this be closed?

gonuke:
superseded by #129

PR #109, opened 2024-05-31
Fix NWL plot

connoramoreno:
The matrix returned by `numpy.historgram2d` in `nwl_utils.py` has rows corresponding to our toroidal angles and columns corresponding to our poloidal angles. However, when we plot the NWL, we set our x-axis to our toroidal angles and our y-axis to our poloidal angles, which is the reverse of the NWL matrix. This PR transposes the NWL matrix such that its rows and columns correspond to the x- and y-axes appropriately.


gonuke (approved):
LGTM

Issue #110, opened 2024-06-01
Passing plasma_mat_tag and sol_mat_tag with split chamber false results in an error


Edgar-21:
I think we should probably just catch this and provide a useful error message?

connoramoreno:
It should result in an error, since `split_chamber` determines whether plasma and SOL components are generated; the same goes for passing `chamber_mat_tag` when `split_chamber` is `True`. My thought was that the error message that results naturally would be enough, and that specially catching it would not be necessary, but perhaps the current error workflow isn't as self-explanatory as I think.

Edgar-21:
It took me a while to figure it out, I'd imagine it would be even more cryptic without some level of familiarity with parastell

gonuke:
Do we need to improve documentation for this? error messages? or just close it?

connoramoreno:
I think we should at least incorporate a specialized error message for this, perhaps in the corresponding setter/getter.

PR #111, opened 2024-06-01
add option to export to cub5

### LINKED ISSUE #93: Add an option to save a Cubit `cub` file
When building a model, it can be useful to save a native Cubit file to facilitate additional operations outside of parastell.

* [x] add an option to argparse and/or YAML
* [x] implement the code to execute this option

Q: Does the user get to specify the file name?

Edgar-21:
added an option to export to cub5 as an optional arg for the `stellarator.export_dagmc()` method. Added a test for this method as well, not sure if it is necessary or not since this is sort of an internal cubit thing and if the DAGMC file gets generated I would expect that the .cub5 file will also be written, though I suppose this test would catch if coreform ever changes that syntax on us.

Also added a small change to config.yaml to allow it to work from the command line, I tested both from the command line and from the python interface and things seem to work as expected. Tests are passing locally as well.

closes #93
works with yaml, and has the same filename as specified for the DAGMC file

### DIFF
--- a/Examples/config.yaml
+++ b/Examples/config.yaml
@@ -63,6 +63,7 @@ invessel_build:
       mat_tag: vac_vessel
   plasma_mat_tag: Vacuum
   sol_mat_tag: Vacuum
+  split_chamber: True
   export_cad_to_dagmc: False
 
 magnet_coils:
@@ -82,4 +83,5 @@ source_mesh:
 dagmc_export:
   skip_imprint: False
   legacy_faceting: True
+  export_cub5: False
   filename: dagmc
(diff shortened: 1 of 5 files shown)


Edgar-21:
I see I am not passing CI tests I'll try to get to the bottom of it

Examples/config.yaml
```diff
@@ -63,6 +63,7 @@ invessel_build:
       mat_tag: vac_vessel
   plasma_mat_tag: Vacuum
   sol_mat_tag: Vacuum
+  split_chamber: True
```
gonuke, line 66: Is this a necessary part of this PR? or just left over from your testing?

gonuke (changes requested):
One quick comment here.

I think all PRs need to be on branches on the upstream repo and not on forks for CI to work/have access to secrets.

Examples/config.yaml
gonuke: Just saw #110...

Edgar-21:
I'll try from a branch on the upstream

PR #113, opened 2024-06-02
Update syntax to union repeated component segments

connoramoreno:
At some point, CadQuery's syntax to union solids together changed from `Component.union(...)` to `Component.fuse(...)`. This PR updates ParaStell's syntax to reflect that change.


gonuke (approved):
Thanks for this quick fix @connoramoreno

PR #114, opened 2024-06-21
resolve missing library when building dockerfile

### LINKED ISSUE #115: Build & publish Docker image for ParaStell CI fails, unable to find libtiff5

Edgar-21:
The Build & publish Docker image for ParaStell CI action currently is failing since `apt-get` cannot find libtiff5. I noticed the base image had updated just prior to this failure, and tried with tag for the previous docker image which resolved the missing library issue.

Fixes #115

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,28 +7,28 @@ RUN apt-get upgrade -y
 
 # Install dependencies
 RUN apt-get install -y libgl1-mesa-glx \
-    libgl1-mesa-dev \
-    libglu1-mesa-dev \
-    freeglut3-dev \
-    libosmesa6 \
-    libosmesa6-dev \
-    libgles2-mesa-dev \
-    curl \
-    wget \
-    libx11-6 \
-    libxt6 \
-    libgl1 \
-    libxcb-icccm4 \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -7,28 +7,28 @@ RUN apt-get upgrade -y
 
 # Install dependencies
 RUN apt-get install -y libgl1-mesa-glx \
-                        libgl1-mesa-dev \
```
gonuke, line 10: original indentation is a little clearer as these are arguments to `apt-get`

gonuke (changes requested):
Can you saw more in the description about which library was missing and/or refer to an issue that documents this?  It's useful to know what problem a PR is actually solving.

Edgar-21:
My apologies, I edited the description to provide some more detail. This was mentioned briefly in the parastell channel a few weeks back as well, just took me a while to get around to submitting this PR.

gonuke:
It looks like `libtiff5-dev` is available - can we try switching to that and unpinning the miniconda image version?  Pinning specific versions usually ends up leading to trouble down the road...

gonuke (approved):

PR #117, opened 2024-06-24
Fix get_loci() method of Surface, so it returns a list of points rather than a list of methods.

### LINKED ISSUE #116: Surface.get_loci() doesn't return rib loci
https://github.com/svalinn/parastell/blob/1e26b751345ecc108facad843e3f255f0602cb97/parastell/invessel_build.py#L413

I suspect it is due to the parenthesis on `rib.rib_loci()` in the line linked above. rib_loci is a parameter, not a method.

Edgar-21:
Closes #116, I also added a line to `test_invessel_build.py` to extend coverage to this method, seemed like a good thing to do, lmk if that is not a best practice.

### DIFF
--- a/tests/test_invessel_build.py
+++ b/tests/test_invessel_build.py
@@ -96,6 +96,7 @@ def test_ivb_basics(invessel_build):
     assert invessel_build.repeat == repeat_exp
     assert invessel_build.num_ribs == num_ribs_exp
     assert invessel_build.num_rib_pts == num_rib_pts_exp
+    assert len(invessel_build.Surfaces['component'].get_loci()) == num_rib_pts_exp
     assert invessel_build.scale == scale_exp
 
     remove_files()
--- a/tests/test_invessel_build.py
+++ b/tests/test_invessel_build.py
@@ -96,7 +96,6 @@ def test_ivb_basics(invessel_build):
     assert invessel_build.repeat == repeat_exp
     assert invessel_build.num_ribs == num_ribs_exp
     assert invessel_build.num_rib_pts == num_rib_pts_exp
-    assert len(invessel_build.Surfaces['component'].get_loci()) == num_rib_pts_exp
     assert invessel_build.scale == scale_exp
 
     remove_files()
@@ -105,6 +104,7 @@ def test_ivb_basics(invessel_build):
 def test_ivb_construction(invessel_build):
 
     num_components_exp = 2
+    len_loci_pt_exp = 3
 
     remove_files()
 
@@ -112,7 +112,11 @@ def test_ivb_construction(invessel_build):
     invessel_build.calculate_loci()
     invessel_build.generate_components()
 
+    rib_loci = invessel_build.Surfaces['component'].get_loci()[0]
+    
+
     assert len(invessel_build.Components) == num_components_exp
+    assert len(rib_loci[0]) == len_loci_pt_exp
 
     remove_files()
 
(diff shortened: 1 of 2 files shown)


tests/test_invessel_build.py
```diff
@@ -96,6 +96,7 @@ def test_ivb_basics(invessel_build):
     assert invessel_build.repeat == repeat_exp
     assert invessel_build.num_ribs == num_ribs_exp
     assert invessel_build.num_rib_pts == num_rib_pts_exp
+    assert len(invessel_build.Surfaces['component'].get_loci()) == num_rib_pts_exp
```
gonuke, line 99: I think we need to call `invessel_build.calculate_loci()` before we can `get` them

gonuke: Or this new test should come as part of the next test where we already call `calculate_loci()`

gonuke (changes requested):
Good catch @Edgar-21 - I think the test needs some tweaking and then it should be fine.

tests/test_invessel_build.py
Edgar-21: sweet, I moved it there and then updated the method a bit as I realized as written here this test actually wouldn't have caught this issue.

gonuke (approved):

tests/test_invessel_build.py
```diff
@@ -104,14 +104,19 @@ def test_ivb_basics(invessel_build):
...
     num_components_exp = 2
+    len_loci_pt_exp = 3
 
     remove_files()
 
     invessel_build.populate_surfaces()
     invessel_build.calculate_loci()
     invessel_build.generate_components()
 
+    rib_loci = invessel_build.Surfaces['component'].get_loci()[0]
```
gonuke, line 115: should we also test `ivb.get_loci()`?

gonuke (changes requested):
One question about expanding testing

tests/test_invessel_build.py
```diff
@@ -104,14 +104,19 @@ def test_ivb_basics(invessel_build):
...
+    
+
     assert len(invessel_build.Components) == num_components_exp
+    assert len(rib_loci[0]) == len_loci_pt_exp
```
gonuke, line 119: Why do we expect only 3 pts? I'm not sure I understand the shape here?

Is this basically `len(poloidal_angles_exp) - 1`

tests/test_invessel_build.py
Edgar-21: `ivb.get_loci()` just loops over the surfaces and uses the `Surface.get_loci()` method, I figured that it would be ok to test just at one surface

tests/test_invessel_build.py
Edgar-21: I expect an element of `rib_loci` to have an iterable of (x,y,z).

I think perhaps I am too focused on testing for the issue I set out to address in the first place, maybe a better testing method would be to have expected values for x,y,z for a specific point on a specific rib and check those?

For example call `ivb.get_loci()` or `Surface.get_loci()`, then get the loci for a single rib, and from that a single loci to check? This would not only check that the `get_loci()` methods are correct, but that the points being calculated are also not changing. I think there might be floating point error to worry about with that sort of comparison though?

gonuke: I think we could at least test for the number of points on the rib? That would probably be enough

Edgar-21: that's getting tested in line 98 already, and didn't catch that loci weren't being returned since the length was as expected, it was just filled with methods rather than loci.

Edgar-21:
added a check to assert that the elements of `rib_loci` are float

gonuke (approved):
LGTM - thanks @Edgar-21

PR #119, opened 2024-06-25
change default value to match docstring in ivb

### LINKED ISSUE #118: mismatch between docstring and default values
in https://github.com/svalinn/parastell/blob/002adc28f47952ad31ed29c19843084d37aae884/parastell/invessel_build.py#L101
a default value of 67 is assigned to `self.num_rib_pts`, the docstring says it should be 61.

Edgar-21:
fixes #118 

@connoramoreno I just guessed that this should be updated to match the docstring rather than the other way around, lmk if that is not the case

### DIFF
--- a/parastell/invessel_build.py
+++ b/parastell/invessel_build.py
@@ -98,7 +98,7 @@ class InVesselBuild(object):
         
         self.repeat = 0
         self.num_ribs = 61
-        self.num_rib_pts = 67
+        self.num_rib_pts = 61
         self.scale = m2cm
 
         for name in kwargs.keys() & (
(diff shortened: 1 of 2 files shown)


connoramoreno:
Let's keep the default value at 67 and just update the docstring. Good catch!

Edgar-21:
black autoformatting got to one of these files hence the big diff

Edgar-21:
CI seems to be failing on `test_magnet_coils.py` for some reason? I didn't touch anything to do with that in this pr

Edgar-21:
nevermind its because this is on my fork, I'll have to reopen a different pr doing the same thing

Issue #120, opened 2024-06-26
NWL calculation should be parallelized


Edgar-21:
The root finding method takes the overwhelming majority of the time required to calculate the NWL map, since this is a repeated calculation in a for loop I believe this is a good candidate for some simple multithreading

Edgar-21:
closed via #122

PR #122, opened 2024-06-26
Parallelize nwl

Edgar-21:
addresses #120

### DIFF
--- a/parastell/nwl_utils.py
+++ b/parastell/nwl_utils.py
@@ -140,7 +140,6 @@ def find_coords(vmec, wall_s, phi, pt):
 
 
 def find_coord(data):
-    print("pid ", os.getpid())
     thetas = []
     vmec = read_vmec.VMECData(data[0])
     for t in data[2]:
@@ -166,7 +165,7 @@ def flux_coords(plas_eq, wall_s, coords):
 
     phi_coords = np.arctan2(coords[:, 1], coords[:, 0])
 
-    num_chunks = 5
+    num_chunks = 10
 
     chunk_size = len(phi_coords) // num_chunks
 
@@ -187,7 +186,7 @@ def flux_coords(plas_eq, wall_s, coords):
             list(executor.map(find_coord, chunks))
         ).flatten()
 
-    return phi_coords, theta_coords
+    return phi_coords.tolist(), theta_coords.tolist()
 
     # phi_coords = np.arctan2(coords[:, 1], coords[:, 0])
     # theta_coords = []
@@ -289,6 +288,7 @@ def nwl_plot(
     num_theta=101,
     num_levels=10,
     num_crossings=None,
+    step_size = 1_000_000
 ):
     """Computes and plots NWL. Assumes toroidal extent is less than 360 degrees
 
@@ -321,7 +321,16 @@ def nwl_plot(
 
     vmec = read_vmec.VMECData(plas_eq)
 
-    phi_coords, theta_coords = flux_coords(plas_eq, wall_s, coords)
+    phi_coords = []
+    theta_coords = []
+
+    iterations = len(coords)//step_size
+
+    for i in range(iterations):
+        print(i, ' million')
+        phi_coord_subset, theta_coord_subset = flux_coords(plas_eq, wall_s, coords[i*step_size:(i+1)*step_size])
+        phi_coords += phi_coord_subset
+        theta_coords += theta_coord_subset
 
     # Define minimum and maximum bin edges for each dimension
     phi_min = 0 - tor_ext / num_phi / 2
--- a/parastell/nwl_utils.py
+++ b/parastell/nwl_utils.py
@@ -324,6 +324,8 @@ def nwl_plot(
 
     if chunk_size is not None:
         iterations = math.ceil(len(coords) / chunk_size)
+    else:
+        chunk_size = len(coords)
 
     for i in range(iterations):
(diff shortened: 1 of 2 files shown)


parastell/nwl_utils.py
```diff
@@ -159,14 +160,27 @@ def flux_coords(vmec, wall_s, coords):
...
     """
-    phi_coords = np.arctan2(coords[:,1], coords[:,0])
-    theta_coords = []
-    
-    for pt, phi in zip(coords, phi_coords):
-        theta = find_coords(vmec, wall_s, phi, pt)
-        theta_coords.append(theta)
 
-    return phi_coords, theta_coords
+    phi_coords = np.arctan2(coords[:, 1], coords[:, 0])
```
Edgar-21, line 164: netcdf4 datasets do not play nicely with some built in functions that would otherwise be nice for this. What happens here is that the particles are split up according to the number of threads available, then each individual process will load its own instance of the vmec object, loop through the particles assigned, then the results are combined here.

parastell/nwl_utils.py
```diff
@@ -112,42 +109,46 @@ def min_problem(theta, vmec, wall_s, phi, pt):
...
-    
+    fw_pt = fw_pt * m2cm
+
     diff = np.linalg.norm(pt - fw_pt)
 
     return diff
 
 
-def find_coords(vmec, wall_s, phi, pt):
+def find_coords(data):
```
Edgar-21, line 119: this function takes a tuple so it plays nicely with `concurrent.futures.ProcessPoolExecutor`

parastell/nwl_utils.py
gonuke: So your saying we can't load the vmec object once and then pass it around?

gonuke (commented):
We'll have to get everyone on the same PEP8 autoformatting to stop thrashing between line length and quotation styles :)

connoramoreno:
I have autopep8 installed, not sure if others would prefer something else.

Edgar-21:
I changed to black since autopep8 wasn't being aggressive enough for the way I write code apparently  🙃. It does a better job of enforcing quotations (and stops me from putting closing brackets on the line I'm not supposed to)

parastell/nwl_utils.py
Edgar-21: Correct. From what I understand, multiprocessing uses pickle, and netcdf4 doesn't pickle apparently, so it can't be passed around by multiprocess. I tried originally to use `multiprocessing.starmap` (which would have been awesome because nothing else would have needed to change) and ran into this issue. Several different forum posts suggested chunking the for loop and loading the data separately for each chunk so I went with that.

Edgar-21:
To test that this parallelization is working as expected I ran 10000 particles from a surface source using the main branch, and wrote the resulting matrix to file, then did the same but with this branch, using 9 threads. I then loaded the files to compare the values and found that they were identical. Not a very formal test, but I could make a PR adding testing for nwl_utils then once that's good to go this PR would be tested against that?

parastell/nwl_utils.py
```diff
@@ -159,14 +161,27 @@ def flux_coords(vmec, wall_s, coords):
...
+        )
+        chunks.append((plas_eq, wall_s, chunk))
+
+    with concurrent.futures.ProcessPoolExecutor(
+        max_workers=num_threads
+    ) as executor:
+        theta_coords = list(executor.map(find_coords, chunks))
+        theta_coords = [
+            theta_coord for chunk in theta_coords for theta_coord in chunk
+        ]
```
gonuke, line 183: Making sure I understand with new variable names
```suggestion
        theta_coord_chunks = list(executor.map(find_coords, chunks))
        theta_coords = [
            theta_coord for theta_coord_chunk in theta_coord_chunks for theta_coord in theta_coord_chunk
        ]
```

parastell/nwl_utils.py
```diff
@@ -274,40 +305,61 @@ def nwl_plot(
...
+
     coords = extract_coords(source_file)
 
     if num_crossings is not None:
         coords = coords[0:num_crossings]
-    
-    # Load plasma equilibrium data
+
+    if num_threads is None:
+        num_threads = cpu_count()
```
gonuke, line 315: That's agressive - can we say `cpu_count()/2` or something as the default to avoid taking all the resources?

gonuke: Default could also be 1??

parastell/nwl_utils.py
```diff
@@ -274,40 +305,61 @@ def nwl_plot(
...
+    phi_coords = []
+    theta_coords = []
+
+    # split up the work to avoid memory issues
+    iterations = 1
+
+    if chunk_size is not None:
+        iterations = math.ceil(len(coords) / chunk_size)
+    else:
+        chunk_size = len(coords)
```
gonuke, line 328: Doesn't this do the same thing?
```suggestion
    if chunk_size is None:
        chunk_size = len(coords)

    iterations = math.ceil(len(coords) / chunk_size)
```

gonuke: Why not call this `chunks` instead of `iterations` and then `for chunk in range(chunks):`

gonuke (changes requested):
Thanks for sticking with this @Edgar-21

gonuke (approved):
LGTM - thanks @Edgar-21

PR #124, opened 2024-06-27
Change tuples of `kwargs` to lists

### LINKED ISSUE #123: `scale` not properly set for `SourceMesh`
Supplied `kwargs` are checked against a tuple of allowed strings. `SourceMesh` has only one allowed keyword argument, `scale`. Potentially related to this fact, the tuple format prevents `scale` from being properly set.

connoramoreno:
Changes the tuple of strings against which keyword arguments are checked to a list format.

Fixes #123.

### DIFF
--- a/parastell/source_mesh.py
+++ b/parastell/source_mesh.py
@@ -95,7 +95,7 @@ class SourceMesh(object):
 
         self.scale = m2cm
 
-        for name in kwargs.keys() & ['scale']:
+        for name in kwargs.keys() & ('scale',):
             self.__setattr__(name,kwargs[name])
         
         self.strengths = []
--- a/parastell/source_mesh.py
+++ b/parastell/source_mesh.py
@@ -95,7 +95,7 @@ class SourceMesh(object):
 
         self.scale = m2cm
 
-        for name in kwargs.keys() & ('scale'):
+        for name in kwargs.keys() & ['scale']:
             self.__setattr__(name,kwargs[name])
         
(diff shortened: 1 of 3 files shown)


gonuke:
Since these are all read only, this shouldn't make a difference

parastell/source_mesh.py
```diff
@@ -95,7 +95,7 @@ def __init__(
 
         self.scale = m2cm
 
-        for name in kwargs.keys() & ('scale'):
+        for name in kwargs.keys() & ['scale']:
```
gonuke, line 98: Sometimes single strings are treated as lists... not sure if that's the problem, but if it is, I don't think this will fix it.

One fix is to add a dangling comma after the string, inside the list/tuple

connoramoreno: I can confirm that changing the tuple to a list fixes the issue.

connoramoreno: Adding a trailing comma also works though.

gonuke: I think I like the tuple better because of the way it communicates immutability.

connoramoreno: Fair enough, I've reverted to tuples but used a trailing comma for `SourceMesh`.

gonuke (changes requested):
Thanks for finding/fixing this @Edgar-21 - looks like we need more testing :)

gonuke (approved):

PR #125, opened 2024-07-10
Allow coil filaments to have variable number of points

paulromano:
Currently, the logic for building magnet coils assumes that all filaments in the coils file have an equal number of points. When trying to use parastell on a coils file for W7-X, I found that the filaments don't have the same length (namely, non-planar and planar coils for W7-X have a different number of points). This PR updates the logic around storing filaments data and computing an average radial distance to accommodate having a variable number of points per filament.

### DIFF
--- a/parastell/magnet_coils.py
+++ b/parastell/magnet_coils.py
@@ -216,30 +216,20 @@ class MagnetSet(object):
                 sample_counter += 1
             else:
                 coords.append([x, y, z])
-                filaments.append(coords)
+                filaments.append(np.array(coords))
                 sample_counter = 0
-                coords = []
+                coords.clear()
 
-        self.filaments = np.array(filaments)
+        self.filaments = filaments
 
     def _set_average_radial_distance(self):
         """Computes average radial distance of filament points.
         (Internal function not intended to be called externally)
-
-        Arguments:
-            filaments (np array of list of list of float): list of filament
-                coordinates. Each filament is a list of coordinates.
-
-        Returns:
-            average_radial_distance (float): average radial distance of
-                magnets (cm).
         """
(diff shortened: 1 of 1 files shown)


Edgar-21:
This test is likely failing due to the PR coming from a branch on a fork, rather than this repo. Tests are passing locally for me.

paulromano:
Given that the CI for both this and #126 is failing at the same spot (`test_rectangular_magnets`, which starts off by initializing Cubit), it does seem like it is related to the fact that the PR is coming from a fork. I'm happy to push these branches to svalinn if you guys are OK with that (not sure who makes that decision @connoramoreno? @gonuke?), although it would be good to figure out how to get PRs from forks working so that other external contributors don't run into the same thing.

connoramoreno:
@paulromano I'm fine with you pushing directly to Svalinn as long as @gonuke doesn't mind. Although I agree that we should get PRs from forks working...

gonuke:
> Given that the CI for both this and #126 is failing at the same spot (`test_rectangular_magnets`, which starts off by initializing Cubit), it does seem like it is related to the fact that the PR is coming from a fork. I'm happy to push these branches to svalinn if you guys are OK with that (not sure who makes that decision @connoramoreno? @gonuke?), although it would be good to figure out how to get PRs from forks working so that other external contributors don't run into the same thing.

The reason we can't PR from a fork is that we have a GITHUB secret for accessing the Cubit license server.  Since we rely on Cubit, it's probably not something we can resolve.  It's easier to grant folks access to our repo than to share that license server key/secret.

gonuke:
I've invited @paulromano to this repo as a maintainer

PR #126, opened 2024-07-10
Assume pystell is installed properly in Python environment

paulromano:
Right now, imports to pystell are of the form `import src.pystell`. This appears to be the case because pystell uses a `src` directory as the root for its Python package. So, if you add the root directory of the pystell git repository to your `PYTHONPATH`, the imports work correctly. However, pystell has a pyproject.toml file that should be used rather than manipulating the `PYTHONPATH` environment variable. Specifically, pystell can be installed by simply running `pip install git+https://github.com/aaroncbader/pystell_uw.git`.


paulromano:
@connoramoreno Could you trigger the CI again? Curious to see if this passes in light of the CI for #125 also failing

paulromano:
Sorry @connoramoreno, can you trigger one more time? :crossed_fingers:

gonuke:
This will also likely need to be PR'd from the upstream repo

PR #127, opened 2024-07-11
Allow coil filaments to have variable number of points

paulromano:
Currently, the logic for building magnet coils assumes that all filaments in the coils file have an equal number of points. When trying to use parastell on a coils file for W7-X, I found that the filaments don't have the same length (namely, non-planar and planar coils for W7-X have a different number of points). This PR updates the logic around storing filaments data and computing an average radial distance to accommodate having a variable number of points per filament.


gonuke (approved):
Thanks @paulromano - this makes sense to me.

PR #128, opened 2024-07-11
Assume pystell is installed properly in Python environment

paulromano:
Right now, imports to pystell are of the form `import src.pystell`. This appears to be the case because pystell uses a `src` directory as the root for its Python package. So, if you add the root directory of the pystell git repository to your `PYTHONPATH`, the imports work correctly. However, pystell has a pyproject.toml file that should be used rather than manipulating the `PYTHONPATH` environment variable. Specifically, pystell can be installed by simply running `pip install git+https://github.com/aaroncbader/pystell_uw.git`.


paulromano:
CI passing on this one now :heavy_check_mark:

Edgar-21:
I didn't realize that pip could be used like that, nice addition. It got me thinking that it should be possible to just include pystell in the `environment.yml` file, tested it out and seems to be working. I PR'ed ir to this branch here #131

gonuke (approved):
Thanks @paulromano (& @Edgar-21 ) for making this all cleaner

Issue #51, opened 2024-03-04
Facilitate user defined reaction rates


gonuke:
Source mesh currently has hard-coded reaction rates that assume:
* relationship between ion temperature and plasma parameter s
* relationship between ion density and plasma parameter s
* DT fusion reaction rates based on ion temperature and density

Users need to provide a mechanism for defining their own relationship between ion temperature and density as a function of the plasma parameter, s.  Users may also want to use a different fusion reaction rate formula, either by selecting the fuel and we provide some standard rate as a function of T, n, or an even more flexible approach.

connoramoreno:
Has this been addressed?

gonuke:
Closed by #129

PR #129, opened 2024-07-11
Add ability to provide custom source strength functions

Edgar-21:
This PR will allow users to pass custom functions for plasma conditions and reaction rates as keyword arguments to either `Stellarator` or `SourceMesh`. The default behavior is to use the same plasma condition/reaction rate that it always has.

The plasma conditions function takes only one input, the plasma parameter s, and must return whatever inputs are required by the reaction rate function. The reaction rate function should return a single value.

Here's an example where the source strengths follow a sine profile:
image

Should I extend the parastell_example.py script to include an example of creating a custom source definition as well?

Addresses #51

### DIFF
--- a/parastell/source_mesh.py
+++ b/parastell/source_mesh.py
@@ -100,7 +100,7 @@ class SourceMesh(object):
             the reaction_rate() function. Defaults to 
             default_plasma_conditions()
         reaction_rate (function): function that takes the values returned by
-            plasma_conditions() and returns a reaction rate in 
+            plasma_conditions() and returns a reaction rate in reactions/cm3/s
     """
 
     def __init__(
--- a/parastell/source_mesh.py
+++ b/parastell/source_mesh.py
@@ -12,8 +12,10 @@ export_allowed_kwargs = ['filename']
 
 
 def default_reaction_rate(n_i, T_i):
-    """Default reaction rate formula for DT fusion
-    assumes an equal mixture of D and T in a hot plasma.
+    """Default reaction rate formula for DT fusion assumes an equal mixture of 
+    D and T in a hot plasma. From A. Bader et al 2021 Nucl. Fusion 61 116060 
+    DOI 10.1088/1741-4326/ac2991
+
 
     Arguments:
         n_i (float) : ion density (ions per m3)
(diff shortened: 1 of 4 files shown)


parastell/source_mesh.py
```diff
@@ -6,36 +6,56 @@
...
+from .utils import read_yaml_config, filter_kwargs, m2cm, m3tocm3
 
 export_allowed_kwargs = ['filename']
 
 
-def rxn_rate(s):
-    """Calculates fusion reaction rate in plasma.
+def default_reaction_rate(n_i, T_i):
+    """Default reaction rate formula for DT fusion
+    assumes an equal mixture of D and T in a hot plasma.
```
gonuke, line 16: Do we have a reference for this?

parastell/source_mesh.py
```diff
@@ -6,36 +6,56 @@
...
+        * T_i ** (-2 / 3)
+        * np.exp(-19.94 * T_i ** (-1 / 3))
+    )
+
+    return rr / m3tocm3
+
+
+def default_plasma_conditions(s):
+    """Calculates ion density and temperature as a function of the
+    plasma paramter s using assumptions of the ARIES design.
```
gonuke, line 42: I wonder if we have a reference for this?

parastell/utils.py
```diff
@@ -4,7 +4,7 @@
 import numpy as np
 
 m2cm = 100
-
+m3tocm3 = 1e6
```
gonuke, line 7:
```suggestion
m3tocm3 = m2cm * m2cm * m2cm
```

gonuke (changes requested):
Thanks for seeing this to completion @Edgar-21 - I have a couple of small comments

parastell/source_mesh.py
connoramoreno: See [this paper](https://iopscience.iop.org/article/10.1088/1741-4326/ac2991/meta) by Aaron, on page four. It gives this equation using separate functions for D and T concentrations, but here we assume them to be equal (per Aaron's suggestion).

gonuke: That appears to be based on [this paper](https://www.tandfonline.com/doi/abs/10.13182/FST08-A1899).  Can we confirm what they actually use here and update accordingly?

gonuke (approved):
Thanks @Edgar-21

PR #134, opened 2024-08-04
Add check for black autoformatting & format all files

Edgar-21:
Two things.

1. `black --line-length 79  .` was run to format all python files with black, and with a line length of 79 characters, per PEP8. Black by default uses 88.
2. An action that checks that code has been formatted such that applying `black --line-length 79  .` does not result in any changes has been added. It runs on pushes, PRs, and manaully.

When/if a `pyproject.toml` is authored it should include black args, and the action should be updated to reference `pyproject.toml`, rather than including the necessary args in the yaml file.

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,0 +1,26 @@
+name: black
+on: [push, pull_request, workflow_dispatch]
+jobs:
+  black:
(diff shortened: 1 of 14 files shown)


Edgar-21:
PR'ed from the wrong branch

PR #135, opened 2024-08-04
Black autoformatting

Edgar-21:
Two things.

1. `black --line-length 79  .` was run to format all python files with black, and with a line length of 79 characters, per PEP8. Black by default uses 88.
2. An action that checks that code has been formatted such that applying `black --line-length 79  .` does not result in any changes has been added. It runs on pushes, PRs, and manaully.

When/if a `pyproject.toml` is authored it should include black args, and the action should be updated to reference `pyproject.toml`, rather than including the necessary args in the yaml file.

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -19,8 +19,4 @@ jobs:
       
       - name: Run Black check
         run: >
-          black --check --verbose --line-length 79 
-          ./Examples 
-          ./executables 
-          ./parastell 
-          ./tests
+          black --check --verbose --line-length 79 .
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,0 +1,26 @@
+name: black
+on: [push, pull_request, workflow_dispatch]
+jobs:
+  black:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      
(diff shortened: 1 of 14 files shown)


gonuke (approved):
Thanks @Edgar-21 - hopefully this helps us avoid too much whitespace churn in reviews.

.github/workflows/black.yml
```diff
@@ -0,0 +1,22 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+      
+      - name: Install Black
+        run: |
+          python -m pip install --upgrade pip
+          pip install black
+      
+      - name: Run Black check
```
gonuke, line 20: Why not use the action provided by `black`: https://black.readthedocs.io/en/stable/integrations/github_actions.html

gonuke (changes requested):
Just found this action provided by PSF that might make life even easier

.github/workflows/black.yml
Edgar-21: much nicer, thank you

Edgar-21: actually not sure if the change I just made is exactly what you had in mind, this is all relatively mysterious to me still

gonuke (approved):
Thanks for updating to the action - let's see if this does what we need.

PR #139, opened 2024-08-27
Change magnet modeling workflow to use CadQuery

### LINKED ISSUE #138: Model magnets using non-licensed software
To improve the ability of the open-source, non-licensed workflow, it would be beneficial to model the magnets using a non-licensed software such as CadQuery.

connoramoreno:
Models the magnets using CadQuery instead of Coreform Cubit. Furthermore, restricts magnet cross-section to rectangular, removing the option for circular magnets.

Closes #138

### DIFF
--- a/parastell/magnet_coils.py
+++ b/parastell/magnet_coils.py
@@ -69,7 +69,13 @@ class MagnetSet(object):
     """
 
     def __init__(
-        self, coils_file, width, thickness, toroidal_extent, logger=None, **kwargs
+        self,
+        coils_file,
+        width,
+        thickness,
+        toroidal_extent,
+        logger=None,
+        **kwargs,
     ):
 
         self.logger = logger
@@ -237,7 +243,9 @@ class MagnetSet(object):
         lower_bound = 2 * np.pi - tol
         upper_bound = self._toroidal_extent + tol
 
-        for coords, tangents in zip(self.filament_coords, self.filament_tangents):
+        for coords, tangents in zip(
+            self.filament_coords, self.filament_tangents
+        ):
             # Compute filament center of mass
             com = np.average(coords, axis=0)
             # Compute toroidal angle of each point in filament
@@ -308,7 +316,9 @@ class MagnetSet(object):
         self._filter_filaments()
 
         self.magnet_coils = [
-            MagnetCoil(coords, tangents, center_of_mass, self._width, self._thickness)
+            MagnetCoil(
+                coords, tangents, center_of_mass, self._width, self._thickness
+            )
             for coords, tangents, center_of_mass in zip(
                 self.filament_coords, self.filament_tangents, self.filament_com
             )
@@ -332,18 +342,22 @@ class MagnetSet(object):
         self.export_dir = export_dir
         self.step_filename = step_filename
 
-        export_path = Path(self.export_dir) / Path(self.step_filename).with_suffix(
-            ".step"
-        )
+        export_path = Path(self.export_dir) / Path(
+            self.step_filename
+        ).with_suffix(".step")
 
-        coil_set = cq.Compound.makeCompound([coil.solid for coil in self.magnet_coils])
+        coil_set = cq.Compound.makeCompound(
+            [coil.solid for coil in self.magnet_coils]
+        )
         cq.exporters.export(coil_set, str(export_path))
 
     def mesh_magnets(self):
         """Creates tetrahedral mesh of magnet volumes via Coreform Cubit."""
         self._logger.info("Generating tetrahedral mesh of magnet coils...")
 
-        last_vol_id = cubit_io.import_step_cubit(self.step_filename, self.export_dir)
+        last_vol_id = cubit_io.import_step_cubit(
+            self.step_filename, self.export_dir
+        )
 
         self.volume_ids = range(1, last_vol_id + 1)
 
@@ -363,7 +377,9 @@ class MagnetSet(object):
         """
         self._logger.info("Exporting mesh H5M file for magnet coils...")
 
-        cubit_io.export_mesh_cubit(filename=mesh_filename, export_dir=export_dir)
+        cubit_io.export_mesh_cubit(
+            filename=mesh_filename, export_dir=export_dir
+        )
 
 
 class MagnetCoil(object):
@@ -395,12 +411,16 @@ class MagnetCoil(object):
         Returns:
             coil (object): cq.Solid object representing a single magnet coil.
         """
-        tangent_vectors = [cq.Vector(tuple(tangent)) for tangent in self.tangents]
+        tangent_vectors = [
+            cq.Vector(tuple(tangent)) for tangent in self.tangents
+        ]
 
         # Define coil filament path normals such that they face the filament
         # center of mass
         normal_dirs = np.array([i - self.center_of_mass for i in self.coords])
-        normal_dirs = normal_dirs / np.linalg.norm(normal_dirs, axis=1)[:, np.newaxis]
+        normal_dirs = (
+            normal_dirs / np.linalg.norm(normal_dirs, axis=1)[:, np.newaxis]
+        )
 
         # Project normal directions onto desired coil cross-section (CS) plane
         # at each filament position to define true filament path normals
@@ -419,26 +439,44 @@ class MagnetCoil(object):
         coil_edge_coords = []
 
         coil_edge1_coords = (
-            self.coords - (self.width / 2) * binormals - (self.thickness / 2) * normals
+            self.coords
+            - (self.width / 2) * binormals
+            - (self.thickness / 2) * normals
         )
-        coil_edge_coords.append([cq.Vector(tuple(pos)) for pos in coil_edge1_coords])
+        coil_edge_coords.append(
+            [cq.Vector(tuple(pos)) for pos in coil_edge1_coords]
+        )
 
         coil_edge2_coords = (
-            self.coords - (self.width / 2) * binormals + (self.thickness / 2) * normals
+            self.coords
+            - (self.width / 2) * binormals
+            + (self.thickness / 2) * normals
         )
-        coil_edge_coords.append([cq.Vector(tuple(pos)) for pos in coil_edge2_coords])
+        coil_edge_coords.append(
+            [cq.Vector(tuple(pos)) for pos in coil_edge2_coords]
+        )
 
         coil_edge3_coords = (
-            self.coords + (self.width / 2) * binormals + (self.thickness / 2) * normals
+            self.coords
+            + (self.width / 2) * binormals
+            + (self.thickness / 2) * normals
         )
-        coil_edge_coords.append([cq.Vector(tuple(pos)) for pos in coil_edge3_coords])
+        coil_edge_coords.append(
+            [cq.Vector(tuple(pos)) for pos in coil_edge3_coords]
+        )
 
         coil_edge4_coords = (
-            self.coords + (self.width / 2) * binormals - (self.thickness / 2) * normals
+            self.coords
+            + (self.width / 2) * binormals
+            - (self.thickness / 2) * normals
         )
-        coil_edge_coords.append([cq.Vector(tuple(pos)) for pos in coil_edge4_coords])
+        coil_edge_coords.append(
+            [cq.Vector(tuple(pos)) for pos in coil_edge4_coords]
+        )
         # Append first edge once again
-        coil_edge_coords.append([cq.Vector(tuple(pos)) for pos in coil_edge1_coords])
+        coil_edge_coords.append(
+            [cq.Vector(tuple(pos)) for pos in coil_edge1_coords]
+        )
 
         coil_edges = [
             cq.Edge.makeSpline(coord_vectors, tangents=tangent_vectors).close()
--- a/parastell/magnet_coils.py
+++ b/parastell/magnet_coils.py
@@ -383,11 +383,6 @@ class MagnetCoil(object):
 
         # Project outward directions onto desired coil cross-section (CS) plane
         # at each filament position to define filament path normals
-        parallel_parts = []
-        for dir, tangent in zip(outward_dirs, self.tangents):
-            parallel_parts.append(np.dot(dir, tangent))
-        parallel_parts = np.array(parallel_parts)
-
         parallel_parts = np.diagonal(
             np.matmul(outward_dirs, self.tangents.transpose())
         )
--- a/parastell/magnet_coils.py
+++ b/parastell/magnet_coils.py
@@ -28,16 +28,10 @@ def compute_tangent(prev_line, next_line):
         tangent (array of float): tangent vector at filament point.
     """
     prev_columns = prev_line.strip().split()
-    prev_x = float(prev_columns[0])
-    prev_y = float(prev_columns[1])
-    prev_z = float(prev_columns[2])
-    prev_pt = np.array([prev_x, prev_y, prev_z])
+    prev_pt = np.array([float(coord) for coord in prev_columns[0:3]])
 
     next_columns = next_line.strip().split()
-    next_x = float(next_columns[0])
-    next_y = float(next_columns[1])
-    next_z = float(next_columns[2])
-    next_pt = np.array([next_x, next_y, next_z])
+    next_pt = np.array([float(coord) for coord in next_columns[0:3]])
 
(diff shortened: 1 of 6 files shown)


parastell/magnet_coils.py
```diff
@@ -11,17 +13,46 @@
...
+            filament point next to the current point.
+
+    Returns:
+        tangent (array of float): tangent vector at filament point.
+    """
+    prev_columns = prev_line.strip().split()
+    prev_x = float(prev_columns[0])
+    prev_y = float(prev_columns[1])
+    prev_z = float(prev_columns[2])
+    prev_pt = np.array([prev_x, prev_y, prev_z])
```
gonuke, line 34:
```suggestion
    prev_pt = np.array([float(ord) for ord in prev_columns[0:2]])
```

parastell/magnet_coils.py
```diff
@@ -11,17 +13,46 @@
...
+
+    next_columns = next_line.strip().split()
+    next_x = float(next_columns[0])
+    next_y = float(next_columns[1])
+    next_z = float(next_columns[2])
+    next_pt = np.array([next_x, next_y, next_z])
```
gonuke, line 40:
```suggestion
    next_pt = np.array([float(ord) for ord in next_columns[0:2]])
```

parastell/magnet_coils.py
```diff
@@ -11,17 +13,46 @@
 export_allowed_kwargs = ["step_filename", "export_mesh", "mesh_filename"]
 
 
+def compute_tangent(prev_line, next_line):
+    """Computes tangent at "current" filament point using central difference
```
gonuke, line 17:
```suggestion
    """Computes tangent at "current" filament point using forward difference
```

parastell/magnet_coils.py
```diff
@@ -90,292 +144,223 @@ def logger(self):
...
-            self._logger.error(e.args[0])
-            raise e
-
-        self.shape = shape
-        self.shape_str = shape_str
-        self.mag_len = mag_len
-
-    def _extract_filaments(self):
-        """Extracts filament data from magnet coil data file.
+    def _extract_filament_data(self):
```
gonuke, line 147: This is a good function to review in a software meeting, but we might not want to wait that long.

I think this would be cleaner if you first just converted all the data from the text file into a more convenient numerical format, and then in separate loops or separate functions, downsample the data and compute the tangents.

1. Data file --> list of numpy arrays of coordinates 
2. Downsample numpy arrays, probably with simple array slice notation
3. Compute tangents, ideally in a function that can act on arrays

gonuke: This would also allow you to calculate the tangents at the last minute, rather than carrying them all around until you need them.

parastell/magnet_coils.py
```diff
@@ -90,292 +144,223 @@ def logger(self):
...
+        # Sort filaments by toroidal angle and overwrite respective arrays
+        self.filament_coords = np.array(
+            [x for _, x in sorted(zip(com_toroidal_angles, filtered_coords))]
+        )
+        self.filament_tangents = np.array(
+            [x for _, x in sorted(zip(com_toroidal_angles, filtered_tangents))]
+        )
+        self.filament_com = np.array(
+            [x for _, x in sorted(zip(com_toroidal_angles, com_list))]
+        )
```
gonuke, line 284: What if you calculated these in a setter function for `filament_coords`?

parastell/magnet_coils.py
```diff
@@ -90,292 +144,223 @@ def logger(self):
...
-
-        # Revolve surface to create wedge spanning toroidal extent
-        cubit.cmd(
-            (
-                f"sweep surface {surf_id} zaxis angle "
-                f"{np.rad2deg(self._toroidal_extent)}"
-            )
+        toroidal_region = cq.Workplane("XZ")
+        toroidal_region = toroidal_region.transformed(
+            offset=(1.25 * self.max_radial_distance / 2, 0)
```
gonuke, line 292: It seems useful to assign `1.25 * self.max_radial_distance` to a variable and reuse it throughout

parastell/magnet_coils.py
gonuke: Do you even need to store them, or could you calculate them inside the initializer for `MagnetCoil`?

parastell/magnet_coils.py
```diff
@@ -401,164 +386,110 @@ class MagnetCoil(object):
...
-            f"rotate Surface {surf_id} angle {np.rad2deg(rot_ang_orig)} about "
-            "origin 0 0 0 direction " + " ".join(str(i) for i in norm)
-        )
+        self.coords = coords
+        self.tangents = tangents
+        self.center_of_mass = center_of_mass
+        self.width = width
+        self.thickness = thickness
 
     def create_magnet(self):
```
gonuke, line 408: We need to redo this with everything originally as a numpy array

parastell/magnet_coils.py
```diff
@@ -401,164 +386,110 @@ class MagnetCoil(object):
...
-        t_vec = np.array([1, 0, 0])
+        tangent_vectors = [
+            cq.Vector(tuple(tangent)) for tangent in self.tangents
+        ]
 
-        # Create cross-section for sweep
-        cubit.cmd(f"create surface " + self.shape_str + " zplane")
+        # Define coil filament path normals such that they face the filament
+        # center of mass
+        normal_dirs = np.array([i - self.center_of_mass for i in self.coords])
```
gonuke, line 420: If this is a numpy array, then this line is
```suggestion
        normal_dirs = self.coords - self.center_of_mass
```

parastell/magnet_coils.py
```diff
@@ -401,164 +386,110 @@ class MagnetCoil(object):
...
-        cs_id = cubit.get_last_id("surface")
-        # Cross-section initially populated with normal oriented along z
-        # axis
-        cs_axis = np.array([0, 0, 1])
+        # Project normal directions onto desired coil cross-section (CS) plane
+        # at each filament position to define true filament path normals
+        parallel_parts = []
+        for dir, tangent in zip(normal_dirs, self.tangents):
+            parallel_parts.append(np.dot(dir, tangent))
+        parallel_parts = np.array(parallel_parts)
```
gonuke, line 430: If these are numpy arrays then this block becomes
```suggestion
        parallel_parts = np.dot(normal_dirs, self.tangents)
```
Or close (maybe need to specify an axis?)

parastell/magnet_coils.py
```diff
@@ -401,164 +386,110 @@ class MagnetCoil(object):
...
-        for x, y, z in self.filament:
-            cubit.cmd(f"create vertex {x} {y} {z}")
-            path += [cubit.get_last_id("vertex")]
+        # Compute binormals projected onto CS plane at each position
+        binormals = np.cross(self.tangents, normals)
 
-        # Ensure final vertex in path is the same as the first
-        path += [path[0]]
+        # Compute coordinates of edges of rectangular coils
+        coil_edge_coords = []
```
gonuke, line 439: This can be done in a loop with something like:
```
edge_offsets = np.array([ [-1,-1], [1,-1], [1,1], [-1, 1] ])

for edge_offset in edge_offsets:
    coil_edge = self.coords + 
                edge_offset[0] * binormals * (self.width /2) +
                edge_offset[1] * binormals * (self.thickness /2)

   coil_edge_coords.append(
        [cq.Vector(tuple(pos)) for pos in coil_edge]
        )

```

gonuke (changes requested):
A few comments here, including some data structure design improvements for streamlining the code

parastell/magnet_coils.py
connoramoreno: I believe each of those, except `width` and `thickness` of course, are numpy arrays

parastell/magnet_coils.py
connoramoreno: They are numpy arrays, I just wasn't careful when rewriting from an earlier version.

parastell/magnet_coils.py
connoramoreno: As mentioned above, these are all numpy arrays. Unfortunately, the numpy functions that perform the dot product, `dot` and `inner`, do not have a way of specifying an axis along which to operate. Without a specified axis, they will just perform a full matrix operation.

parastell/magnet_coils.py
connoramoreno: I believe it is central difference since the approximation uses the next and previous points, rather than the forward difference which uses the next and current points.

parastell/magnet_coils.py
connoramoreno: They could be calculated in `MagnetCoil`, but I would want all of the coordinate data available, to better resolve tangents and centers of mass. I think that combines well with your previous comment.

parastell/magnet_coils.py
connoramoreno: My intention was to reduce the total amount of information stored, storing only the sampled points while at the same time computing the tangents using the full set of points. But I agree it makes the code a little confusing, and maybe adds to the total amount of information stored at a given time.

connoramoreno: Also, I was unaware you could use steps when array slicing; very neat!

parastell/magnet_coils.py
gonuke: I see.... but this method actually just uses any two points, and it's your choice of which points to pass as arguments that makes a difference.  So technically, this method simply calculates a unit vector between two points.

parastell/magnet_coils.py
gonuke: I guess my point is that you don't use it until you build the coil.  It could even be temporary data that you generate when building the coil

parastell/magnet_coils.py
gonuke: Can we structure the matrices such that a full matrix operation is the right thing to do?

parastell/magnet_coils.py
connoramoreno: I suppose that's true. That said, the updated version I'm about to commit doesn't rely on this function anymore, so I've removed it from this script.

parastell/magnet_coils.py
connoramoreno: The centers of mass get used to sort the filaments, and unless you think it's worthwhile to recompute them at the coil stage rather than store them here, I've kept it as is. I've moved the computation of the tangents to the coil stage, however.

parastell/magnet_coils.py
```diff
@@ -401,164 +319,104 @@ class MagnetCoil(object):
...
+        )
 
-        # Initialize path list
-        path = []
+        # Project normal directions onto desired coil cross-section (CS) plane
+        # at each filament position to define true filament path normals
+        parallel_parts = []
+        for dir, tangent in zip(normal_dirs, self.tangents):
+            parallel_parts.append(np.dot(dir, tangent))
+        parallel_parts = np.array(parallel_parts)
```
gonuke, line 382: Isn't this the same as
```suggestion
        parallel_parts = np.matmul(normal_dirs, self.tangents.transpose())
```

parastell/magnet_coils.py
```diff
@@ -401,164 +319,104 @@ class MagnetCoil(object):
...
+        ]
 
-        # Store cross-section index
-        cs_id = cubit.get_last_id("surface")
-        # Cross-section initially populated with normal oriented along z
-        # axis
-        cs_axis = np.array([0, 0, 1])
+        # Define coil filament path normals such that they face the filament
+        # center of mass
+        normal_dirs = self._coords - self.center_of_mass
```
gonuke, line 372: There are two variables called `normal` (or similar) but they are different.  Could we use something else for `normal_dirs`, e.g. `radial_dirs` or `outward_dirs`?

gonuke (changes requested):
Looking a lot better... just a few final polishing questions

parastell/magnet_coils.py
```diff
@@ -401,164 +319,104 @@ class MagnetCoil(object):
...
+        tangents = np.subtract(
+            np.append(data[1:], [data[1]], axis=0),
+            np.append([data[-2]], data[0:-1], axis=0),
         )
+        tangents = tangents / np.linalg.norm(tangents, axis=1)[:, np.newaxis]
+
+        # Sample filament coordinates and tangents by modifier
+        self._coords = data[0 : -1 : self.sample_mod]
+        self._coords = np.append(self._coords, [self._coords[0]], axis=0)
+        self.tangents = tangents[0 : -1 : self.sample_mod]
```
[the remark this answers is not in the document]
connoramoreno: Yes, that's what I was going for :)

parastell/magnet_coils.py
```diff
@@ -401,164 +319,104 @@ class MagnetCoil(object):
...
+        return self._coords
+
+    @coords.setter
+    def coords(self, data):
+        # Compute tangents
+        tangents = np.subtract(
+            np.append(data[1:], [data[1]], axis=0),
+            np.append([data[-2]], data[0:-1], axis=0),
         )
+        tangents = tangents / np.linalg.norm(tangents, axis=1)[:, np.newaxis]
```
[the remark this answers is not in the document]
connoramoreno: I think there may be some confusion here. What I wrote computes, at each filament point, the normalized direction from the previous point to the next point. Computation of the filament normals and binormals relies on the tangents being unit vectors. From what I infer, what has been suggested attempts to compute essentially the same thing, except that the tangents are not normalized at the end. I know I mentioned previously that the "central difference approximation" is used to compute tangents, but that term was used loosely to convey which points are used to compute the tangent rather than an actual derivative.

It may also help to remember that for each filament, the initial and final points are the same. So, when computing the tangent for each of those points, we want to compute the difference between the second and second-to-last points in the array. That is why we add the second point to the end of the first array and second-to-last point to the beginning of the second array in the `np.subtract` operation. We are essentially just rearranging the loops, shifting the first and second arrays one index forward and one index backward, respectively.

Adding the first point to the end of the first array and last point to the beginning of the second array, as suggested, will lead to zero vectors computed for the tangents at both the first and final points.

parastell/magnet_coils.py
connoramoreno: Almost. We'd then need to take the diagonal of the resultant matrix (`parallel_parts` is supposed to be an array of scalars and `np.matmul(normal_dirs, self.tangents.transpose())` results in an NxN matrix where N is the number of points on the filament), but that's easy enough to do.

parastell/magnet_coils.py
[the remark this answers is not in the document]
connoramoreno: I see where you're coming from now, I wasn't even thinking of central difference as an average of the forward and backward differences but rather just as the "end product" where the point of interest cancels out; apologies for that. I think the implementation I wrote is straight-forward and does what we want.

parastell/magnet_coils.py
```diff
@@ -401,164 +324,105 @@ class MagnetCoil(object):
...
+            outward_dirs / np.linalg.norm(outward_dirs, axis=1)[:, np.newaxis]
+        )
 
-        # Initialize path list
-        path = []
+        # Project outward directions onto desired coil cross-section (CS) plane
+        # at each filament position to define filament path normals
+        parallel_parts = np.diagonal(
+            np.matmul(outward_dirs, self.tangents.transpose())
+        )
```
gonuke, line 388: See my comment elsewhere... would this get you a vector if you swapped the transpose?
```suggestion
        parallel_parts = p.matmul(outward_dirs.transpose(), self.tangents)
```
They are both 3 x N matrices, right? so there must be some way to arrange them such that a matrix multiplication results in a 1 x N vector without needing to take the diagonal...

gonuke (changes requested):
Still puzzled by the matrix multiply

parastell/magnet_coils.py
connoramoreno: Perhaps I'm not seeing the same thing as you but I don't believe there's a way for a matrix product to be smaller in a dimension than the smallest dimension of the original matrices. I.e., for two Nx3 matrices, we can produce a 3x3 matrix (first matrix transposed) or a NxN matrix (second matrix transposed). That said, there may be some NumPy operation capable of producing what we want but I can't seem to find one...

gonuke: I really am getting old.... and would probably fail my qual...  I'll stop questioning these things when they work...

gonuke (approved):
Sorry for the runaround on all these issues...

PR #142, opened 2024-08-29
Adds utility for building the 'magnet surface' and shooting rays at it

Edgar-21:


parastell/radial_distance_utils.py
```diff
@@ -0,0 +1,171 @@
...
+# from matplotlib import pyplot as plt
+# from matplotlib.ticker import FormatStrFormatter
+from . import magnet_coils
+from . import invessel_build as ivb
+import pystell.read_vmec as read_vmec
+
+import cubit
+
+
+# function to get reordered filaments
```
Edgar-21, line 12:
```suggestion

```

parastell/radial_distance_utils.py
```diff
@@ -0,0 +1,171 @@
...
+def reorder_filaments(filaments):
+    """
+    Reorder the filaments so they start near the outboard xy plane crossing,
+    and begin by increasing z value.
+
+    Arguments:
+        filaments (list of list of list of float): List of filaments, which are
+            lists of points defining each filament.
+    Returns:
+        filaments (list of list of list of float): Reorderd list of filaments,
```
Edgar-21, line 86:
```suggestion
        filaments (list of list of list of float): Reordered list of filaments,
```

Edgar-21:
I think #139 breaks this - but only a little. Once that gets merged I can update this

parastell/radial_distance_utils.py
```diff
@@ -0,0 +1,170 @@
...
+        filaments (list of list of list of float): filaments in order of
+            increasing toroidal angle.
+    """
+    com_list = []
+
+    for fil in filaments:
+        com = np.average(fil, axis=0)
+        com_list.append(com)
+
+    com_list = np.array(com_list)
```
gonuke, line 68:
```suggestion
    com_list = np.zeros(len(filaments))

    for idx, fil in enumerate(filaments):
        com_list[idx] = np.average(fil, axis=0)
```

parastell/radial_distance_utils.py
```diff
@@ -0,0 +1,170 @@
...
+    """
+    for filament_index, filament in enumerate(filaments):
+        # start the filament at the outboard
+        max_z_index = get_start_index(filament)
+        reordered_filament = np.concatenate(
+            [filament[max_z_index:], filament[0:max_z_index]]
+        )
+
+        # make sure z is increasing initially
+        if filament[max_z_index, 2] > filament[max_z_index + 1, 2]:
```
gonuke, line 96:
```suggestion
        if reordered_filament[0, 2] > reordered_filament[1, 2]:
```

parastell/radial_distance_utils.py
```diff
@@ -0,0 +1,170 @@
...
+            reordered_filament = np.flip(reordered_filament, axis=0)
+
+        # remove duplicate point since the start point might be different
+        _, idx = np.unique(reordered_filament, return_index=True, axis=0)
+        reordered_filament = reordered_filament[np.sort(idx)]
```
gonuke, line 101: If the duplicate is always (???) the start point before you reorder it, is it easier to remove it before you reorder it?

parastell/radial_distance_utils.py
```diff
@@ -0,0 +1,170 @@
...
+    filtered_filaments = magnet_set.filtered_filaments
+    filaments = reorder_filaments(filtered_filaments)
+
+    return filaments
+
+
+def build_magnet_surface(reordered_filaments):
+    for fil_index, _ in enumerate(reordered_filaments[0:-1]):
+        fil1 = reordered_filaments[fil_index]
+        fil2 = reordered_filaments[fil_index + 1]
```
gonuke, line 132:
```suggestion
    for fil1, fil2 in zip(reordered_filaments[0:-1], reordered_filaments[1:]):
```

parastell/radial_distance_utils.py
```diff
@@ -0,0 +1,170 @@
...
+            z2 = fil2[index, 2]
+            cubit.cmd(
+                f"create curve location {x1} {y1} {z1} location {x2} {y2} {z2}"
+            )
+
+    lines = np.array(cubit.get_entities("curve"))
+    lines = np.reshape(
+        lines, (len(reordered_filaments) - 1, len(reordered_filaments[0]))
+    )
+    for loop in lines:
```
gonuke, line 148:
```suggestion
    loops = np.reshape(
        lines, (len(reordered_filaments) - 1, len(reordered_filaments[0]))
    )
    for loop in loops:
```

parastell/radial_distance_utils.py
```diff
@@ -0,0 +1,170 @@
...
+            cubit.cmd(f"create surface skin curve {line} {line + 1}")
+
+
+def measure_radial_distance(ribs):
+    distances = []
+    for rib in ribs:
+        distance_subset = []
+        for point, direction in zip(rib.rib_loci, rib._normals()):
+            cubit.cmd(f"create vertex {point[0]} {point[1]} {point[2]}")
+            vertex_id = max(cubit.get_entities("vertex"))
```
gonuke, line 159: Isn't there a better way to get the last vertex than getting all of the vertices and taking the max?

gonuke (changes requested):

parastell/radial_distance_utils.py
gonuke: I think you might be able to use ```vtx_id = cubit.get_last_id("vertex")```

parastell/radial_distance_utils.py
Edgar-21: yes that is easier, thank you

parastell/radial_distance_utils.py
```diff
@@ -0,0 +1,156 @@
...
+            x2 = fil2[index, 0]
+            y1 = fil1[index, 1]
+            y2 = fil2[index, 1]
+            z1 = fil1[index, 2]
+            z2 = fil2[index, 2]
+            cubit.cmd(
+                f"create curve location {x1} {y1} {z1} location {x2} {y2} {z2}"
+            )
+
+    loops = np.array(cubit.get_entities("curve"))
```
gonuke, line 130: Does this get all the curves, including those not generated by you above?  If so, are we better off collecting the curves using `cubit.get_last_id()` as you create them?

gonuke (changes requested):
Just one more place to use get_last_id - maybe...

parastell/radial_distance_utils.py
Edgar-21: It does, and we are

parastell/radial_distance_utils.py
```diff
@@ -0,0 +1,158 @@
...
+        for index, _ in enumerate(fil1):
+            x1 = fil1[index, 0]
+            x2 = fil2[index, 0]
+            y1 = fil1[index, 1]
+            y2 = fil2[index, 1]
+            z1 = fil1[index, 2]
+            z2 = fil2[index, 2]
+            cubit.cmd(
+                f"create curve location {x1} {y1} {z1} location {x2} {y2} {z2}"
+            )
```
gonuke, line 129: How about this?
```suggestion
            loc1 = ' '.join(fil1[index,0:2])
            loc2 = ' '.join(fil2[index,0:2])
            cubit.cmd(
                f"create curve location {loc1} location {loc2}"
            )
```

parastell/radial_distance_utils.py
```diff
@@ -0,0 +1,158 @@
...
+        max_z_index (int): index at which the filament crosses the xy plane
+    """
+    max_z_index = None
+    max_z_radius = 0
+    for index, (point, next_point) in enumerate(
+        zip(filament[0:-1], filament[1:])
+    ):
+        if point[2] / next_point[2] < 0:
+            z_radius = calc_z_radius(point)
+            if max_z_radius < z_radius:
```
gonuke, line 39: What if the max z point isn't where the filament crosses the x-y plane?

gonuke (changes requested):
A few more changes...

parastell/radial_distance_utils.py
Edgar-21: Seems like this is a poor variable name. It's called max_z_radius because its should end up being the point where it crosses the x-y plane with the larger distance to the z axis (has the larger radius if the center is the z axis) of the two crossings, not anything to do with the actual z value of the points. The if statement above checks for a sign change in z between the two points, and only checks for what the z radius is if there is a sign change (crosses the xy plane). I'll rename these to something hopefully a bit clearer.

parastell/radial_distance_utils.py
Edgar-21: need to use list comprehension to make a list of str first but this will work. join() doesn't like to operate on iterables of things that are not str seems like

gonuke (approved):
Thanks for putting this all together @Edgar-21

PR #148, opened 2024-09-09
Use non-interactive backend in nwl utils to avoid matplotlib cubit conflict

Edgar-21:
With the newest version of openmc (0.15.0) the previous conflict with Cubit seems to have been resolved. The remaining conflict with matplotlib can be resolved as in this PR, allowing the full NWL workflow to be performed in one script. In a future PR the nwl_utils example could be updated to be a single script.

### DIFF
--- a/parastell/nwl_utils.py
+++ b/parastell/nwl_utils.py
@@ -5,8 +5,10 @@ import openmc
 import pystell.read_vmec as read_vmec
 import matplotlib.pyplot as plt
 import concurrent.futures
-from multiprocessing import cpu_count
 import math
+import matplotlib
+
+matplotlib.use("agg")
 
 
 def extract_ss(ss_file):
--- a/environment.yml
+++ b/environment.yml
@@ -3,16 +3,17 @@ channels:
   - conda-forge
 dependencies:
   - python=3.11.6
-  - pip=24.0
-  - numpy
+  - pip
+  - numpy=1.26.4
   - scipy
   - scikit-learn
   - cadquery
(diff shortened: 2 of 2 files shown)


connoramoreno:
We should perhaps consider including OpenMC v0.15.0 in the environment YAML file

parastell/nwl_utils.py
```diff
@@ -5,8 +5,10 @@
 import pystell.read_vmec as read_vmec
 import matplotlib.pyplot as plt
 import concurrent.futures
-from multiprocessing import cpu_count
```
connoramoreno, line 8: Why are we removing `multiprocessing`? Wouldn't that remove parallelization from the NWL routine?

connoramoreno (commented):
Just a small question about removing the `multiprocessing` module from the list of imports.

parastell/nwl_utils.py
Edgar-21: `concurrent.futures` is what does the multiprocesssing, I originally attempted to make some guess about how many cpus to use based on `cpu_count` but ended up removing that bit, but this import remained

connoramoreno:
I'm in favor of at least including some mention that OpenMC v0.15.0 is needed to use the NWL utility. I'm not opposed to including it as a hard dependency to prevent user error on that front, although I recognize it complicates the environment and may become a problem in the future, like it was in the past.

connoramoreno:
Looks good, thanks @Edgar-21

PR #151, opened 2024-09-10
Refactor filament data use

### LINKED ISSUE #47: Refactor use of filaments for magnet coils
Instead of reading and storing all the filament data in the `MagnetSet` class, use each filament data to initialize a new `MagnetCoil` object as it becomes available.

Also store all of the filament points in the `MagnetCoil` class and only filter to fewer data points when building the curves.

This will then also require a refactor of how we decide which magnets are in the toroidal extent, perhaps with a method on the `MagnetCoil` class that returns a Boolean if it is/is not.

### LINKED ISSUE #149: Radial distance utility lightly broken by CadQuery magnets PR
As foreseen by @Edgar-21, the radial distance utility does not work with the newest version of the magnet routine. I imagine and am hopeful that adapting `radial_distance_utils.py` and `radial_build_distance_example.py` won't be too difficult.

connoramoreno:
Streamlines and reorganizes use of filament coordinate data as follows:

- Instantiates a new `MagnetCoil` class object as filament coordinate data becomes available in the `MagnetSet._instantiate_coils` method (formerly named `MagnetSet._extract_filament_data`), rather than store all filament data separately (and redundantly) in the `MagnetSet` class.
- For the purpose of filtering coils based on toroidal angle, moves the check of whether each coil lies within the model's toroidal extent to the `check_coil_toroidal_extent` method of the `MagnetCoil` class, rather than the `_filter_coils` method of the `MagnetSet` class (formerly named `MagnetSet._filter_filaments`). The `MagnetCoil._check_coil_toroidal_extent` method is called by the `MagnetSet._filter_coils` method to subsequently perform the actual filtering.
- Modifies storage of filament coordinates in the `MagnetCoil` class such that all coordinates are stored, and only downsampled during actual CAD solid generation. Note that the downsampled coordinates are not stored as part of any class.

Closes #47
Closes #149 (via PR #152)

### DIFF
--- a/Examples/radial_build_distance_example.py
+++ b/Examples/radial_build_distance_example.py
@@ -1,29 +1,34 @@
 from parastell.radial_distance_utils import *
 
-coils_file = "coils_wistelld.txt"
-circ_cross_section = ["circle", 25]
+coils_file = "coils.example"
+width = 40.0
+thickness = 50.0
 toroidal_extent = 90.0
-vmec_file = "plasma_wistelld.nc"
+vmec_file = "wout_vmec.nc"
 vmec = read_vmec.VMECData(vmec_file)
 
 magnet_set = magnet_coils.MagnetSet(
-    coils_file, circ_cross_section, toroidal_extent
+    coils_file, width, thickness, toroidal_extent
 )
 
 reordered_filaments = get_reordered_filaments(magnet_set)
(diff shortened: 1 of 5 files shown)


gonuke (approved):
Thanks @connoramoreno

PR #156, opened 2024-09-15
use UWUW vacuum boundary condition for NWL example

### LINKED ISSUE #155: Triple Counting of Some Neutrons in NWL utils
Increasing the maximum number of surface crossings to be larger than the number of particles transported reveals an issue with the way the bounding surfaces are set up in nwl_transport() currently. For example, if I set the number of particles to transport to be 100,000, then set the maximum particles in the surface source, I wind up with 123,890 particles banked on the surface source. I believe there are two factors leading to triple counting. If the boundaries were correct, I would expect to only bank the number of source particles.

I reckon the largest contributor to the extra sites is particles exiting the inboard side, encountering the periodic boundary, and returning to re-enter the chamber. These particles will be counted once as they exit, then again as they re-enter, and a third time as they exit. A similar case can be imagined where the chamber has a convex profile, as a particle may exit and reenter without encountering a boundary.

The best fix would probably be to assign a vacuum boundary condition to the FW surface in cubit prior to exporting the DAGMC model.

Edgar-21:
Updates the example to apply a vacuum boundary condition prior to DAGMC export. Fixes #155. Tried to make this as minimal of a fix as possible

### DIFF
--- a/Examples/nwl_geom_example.py
+++ b/Examples/nwl_geom_example.py
@@ -1,9 +1,6 @@
 from pathlib import Path
-
-import numpy as np
-
 import parastell.parastell as ps
-
+from parastell.cubit_io import tag_surface_legacy
 
 # Define directory to export all output files to
 export_dir = ""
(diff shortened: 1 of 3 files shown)


gonuke (approved):
This looks good to me - thanks for noticing and fixing this @Edgar-21

PR #157, opened 2024-09-18
Modify README and Add Image of Example ParaStell Model

connoramoreno:
Modifies `README.md` to include an image of an example ParaStell model. Also reorders the installation instructions to include cloning the ParaStell repository before creating a conda environment based on `environment.yml`.


gonuke (approved):
LGTM - thanks @connoramoreno

PR #158, opened 2024-09-19
Overhaul radial distance utility

connoramoreno:
Overhauls the radial distance utility. Changes include:

- Leveraging recent updates to the magnet workflow
- Code restructuring
- General code improvements
- Addition of a function to handle all radial distance utility functionality, removing the need to call several functions when using the utility
- Addition of a function to enforce helical symmetry in a matrix whose rows and columns correspond to regularly spaced angular arrays
- Addition of a function to smooth a matrix using Gaussian filtering without increasing the value of matrix elements


parastell/radial_distance_utils.py
```diff
@@ -1,154 +1,360 @@
...
+    Returns:
+        distance_matrix (2-D np.array of float): matrix of distances between
+            points defining surface.Ribs and magnet surface, along
+            Rib._normals.
     """
-    magnet_set.populate_magnet_coils()
+    distance_matrix = np.array(
+        [
+            [
+                fire_ray(point, distance)
```
Edgar-21, line 224: distance should probably be called direction here

connoramoreno: Correct you are, thank you

parastell/radial_distance_utils.py
```diff
@@ -1,154 +1,360 @@
...
+                for point, direction in zip(rib.rib_loci, rib._normals())
+            ]
+            for rib in surface.Ribs
+        ]
+    )
+
+    return distance_matrix
+
+
+def measure_fw_coils_separation(
```
Edgar-21, line 234: If the first layer in the radial build is called "sol" does it get merged into chamber if split_chamber is false?

connoramoreno: No, the layer would have to be named “chamber”

connoramoreno:
I'm beginning to wonder if it might be worth converting this script to an object-oriented version

parastell/radial_distance_utils.py
```diff
@@ -1,154 +1,368 @@
...
     """
-    com_list = np.zeros((len(filaments), 3))
+    com_list = np.array([coil.center_of_mass for coil in magnet_coils])
+    com_toroidal_angles = np.arctan2(com_list[:, 1], com_list[:, 0])
 
-    for idx, fil in enumerate(filaments):
-        com_list[idx] = np.average(fil, axis=0)
+    magnet_coils = np.array(
+        [x for _, x in sorted(zip(com_toroidal_angles, magnet_coils))]
+    )
```
gonuke, line 88: If you define `MagnetCoil.com_toroidal_angle()`
```suggestion
    magnet_coils =  sorted(magnet_coils, key = lambda x: x.com_toroidal_angle)
```

parastell/radial_distance_utils.py
```diff
@@ -1,154 +1,368 @@
...
+    """
+    cubit_io.init_cubit()
+
+    surface_lines = [
+        [
+            build_line(coord, next_coord)
+            for coord, next_coord in zip(
+                np.concatenate(
+                    [coil.coords[:-1:sample_mod], [coil.coords[0]]]
+                ),
```
gonuke, line 151: What if we had a function `downsample_loop()` that does this concatenation

gonuke (changes requested):

parastell/utils.py
```diff
@@ -1,70 +1,92 @@
...
+        list_exp = np.append(
+            list_exp,
+            np.linspace(entry, next_entry, num=num_new_entries + 1)[:-1],
+        )
 
-def read_yaml_config(filename):
-    """Read YAML file describing ParaStell configuration and extract all data."""
-    with open(filename) as yaml_file:
-        all_data = yaml.safe_load(yaml_file)
+    list_exp = np.append(list_exp, list[-1])
```
gonuke, line 87:
```suggestion
    list_exp = np.append(list_exp, final_entry)
```

parastell/utils.py
```diff
@@ -1,70 +1,92 @@
...
+    extent = final_entry - init_entry
 
-    ang_list_exp = np.append(ang_list_exp, ang_list[-1])
+    avg_diff = extent / (num - 1)
 
-    return ang_list_exp
+    for entry, next_entry in zip(list[:-1], list[1:]):
+        num_new_entries = int(round((next_entry - entry) / avg_diff))
 
+        list_exp = np.append(
```
gonuke, line 82: Subtle feature may warrant a comment
```suggestion
        # don't append the last entry in the created linspace to avoid adding it twice when
        # the next created linspace is appended
        list_exp = np.append(
```

parastell/utils.py
```diff
@@ -101,3 +123,73 @@ def filter_kwargs(
...
+
+    Arguments:
+        list (iterable): closed loop.
+        index (int): list index about which to reorder loop.
+
+    Returns:
+        reordered_loop (iterable): reordered closed loop.
+    """
+    reordered_list = np.concatenate([list[index:], list[1:index]])
+    reordered_list = np.append(reordered_list, [reordered_list[0]], axis=0)
```
gonuke, line 165: Is there a case (edge case?) where this doesn't work?
```suggestion
    reordered_list = np.concatenate([list[index:], list[1:index+1]])
```
maybe if  `index == len(list)`?

parastell/utils.py
gonuke, line 1: Most/many/all of these functions are perfect for tests to make sure they are right and robust to edge cases.

parastell/magnet_coils.py
```diff
@@ -436,6 +409,87 @@ def create_magnet(self):
...
+            outboard_index (int): index of the outboard midplane point.
+        """
+        # Compute radial distance of coordinates from z-axis
+        radii = np.linalg.norm(self.coords[:, :2], axis=1)
+        # Determine whether adjacent points cross the midplane (if so, they will
+        # have opposite signs)
+        midplane_flags = -np.sign(
+            self.coords[:, 2]
+            * np.append(self.coords[1:, 2], self.coords[1, 2])
+        )
```
gonuke, line 465: Maybe this makes things more readable?
```suggestion
        shifted_coords = np.append(self.coords[1:, :], self.coords[1, :])
        midplane_flags = -np.sign(self.coords[:, 2] * shifted_coords[:, 2])
```

parastell/magnet_coils.py
```diff
@@ -436,6 +409,87 @@ def create_magnet(self):
...
+        progress positively or negatively.
+
+        Arguments:
+            positive (bool): progress coordinates in positive direciton
+                (defaults to True). If negative, coordinates will progress in
+                negative direction.
+        """
+        if (positive and self.coords[0, 2] > self.coords[1, 2]) or (
+            not positive and self.coords[0, 2] < self.coords[1, 2]
+        ):
```
gonuke, line 490: Since `positive` is a boolean, can this be just:
```suggestion
        if (positive == (self.coords[0, 2] > self.coords[1, 2]) ):
```

Examples/radial_distance_example.py
```diff
@@ -0,0 +1,113 @@
...
+    width,
+    thickness,
+    sample_mod=1,
+)
+# For matrices defined by angles that are regularly spaced, measurement can
+# result in matrix elements that are close to, but not exactly, helcially
+# symmetric
+available_space = enforce_helical_symmetry(available_space)
+# Smooth matrix
+available_space = smooth_matrix(available_space, 100, 1)
```
gonuke, line 42: I think it makes sense for the example to avoid magic numbers, maybe especially so - so that users can figure out what these are for

Examples/radial_distance_example.py
```diff
@@ -0,0 +1,113 @@
...
+# Define directory to export all output files to
+export_dir = ""
+# Define plasma equilibrium VMEC file
+vmec_file = "wout_vmec.nc"
+
+# Instantiate ParaStell build
+stellarator = ps.Stellarator(vmec_file)
+
+# Define build parameters for in-vessel components
+toroidal_angles = np.linspace(0, 90, num=97)
```
gonuke, line 17: magic number: 97

Examples/radial_distance_example.py
```diff
@@ -0,0 +1,113 @@
...
+# For matrices defined by angles that are regularly spaced, matrix smoothing
+# can result in matrix elements that are close to, but not exactly, helcially
+# symmetric
+available_space = enforce_helical_symmetry(available_space)
+# Modify available space to account for thickness of magnets
+available_space = available_space - max(width, thickness)
+
+# Define a matrix of uniform unit thickness
+uniform_unit_thickness = np.ones(
+    (len(toroidal_angles[::8]), len(poloidal_angles[::8]))
```
gonuke, line 52: why are we downsampling here?  Is is necessary? is it showing a feature? is it arbitrary?

gonuke: Any everything seems to be downsampled each time it's used - why not downsample once here for each relevant data structure?

gonuke (changes requested):
I'm really impressed by the application of best practices here @connoramoreno - nice cleanup!  I have a few more comments to continue the journey...

parastell/magnet_coils.py
connoramoreno: Very nice

parastell/utils.py
connoramoreno: Are you thinking I should make another test script?

connoramoreno: If so, I should perhaps do that in a different PR

parastell/utils.py
connoramoreno: Yes, in that case this would result in an index error

connoramoreno: Actually, this would work even in that edge case since `index+1` would never actually be reached.

parastell/utils.py
gonuke: Thes best time to make tests is in the PR where you add the code...

gonuke: Maybe it's overkill for now...

parastell/utils.py
```diff
@@ -1,70 +1,98 @@
...
-        num_ang (int): number of angles to include in stellarator build.
+        matrix (2-D iterable of float): matrix to be made helically symmetric.
 
     Returns:
-        ang_list_exp (list of double): interpolated list of angles (rad).
+        matrix (2-D iterable of float): helically symmetric matrix.
     """
-    ang_list = np.deg2rad(ang_list)
+    num_rows = matrix.shape[0]
+    num_columns = matrix.shape[1]
```
gonuke, line 35:
```suggestion
    num_rows, num_columns = matrix.shape
```

parastell/utils.py
```diff
@@ -1,70 +1,98 @@
...
 
-    ang_list_exp = []
+    # Ensure rows represent closed loops
+    for idx in range(num_rows):
+        matrix[idx, -1] = matrix[idx, 0]
```
gonuke, line 39:
```suggestion
    matrix[:,-1] = matrix[:,0]
```

parastell/utils.py
```diff
@@ -1,70 +1,98 @@
...
+    # Ensure helical symmetry toroidally and poloidally by mirroring the period
+    # about both matrix axes
+    flattened_matrix = matrix.flatten()
 
-    for ang, next_ang in zip(ang_list[:-1], ang_list[1:]):
-        n_ang = math.ceil((next_ang - ang) / ang_diff_avg)
+    for idx, value in enumerate(
+        flattened_matrix[: int(len(flattened_matrix) / 2)], start=1
+    ):
+        flattened_matrix[-idx] = value
```
gonuke, line 56: Is this what it's doing?
```suggestion
    half_period = flattened_matrix[0:int(len(flattened_matrix) / 2)]
    flattened_matrix = np.concatenate([half_period, np.flip(half_period)])
```

parastell/utils.py
```diff
@@ -101,3 +129,70 @@ def filter_kwargs(
...
+
+    Arguments:
+        matrix (2-D iterable of float): matrix to be smoothed.
+        steps (int): number of smoothing steps.
+        sigma (float): standard deviation for Gaussian kernel.
+
+    Returns:
+        smoothed_matrix (2-D iterable of float): smoothed matrix.
+    """
+    previous_iteration = matrix
```
gonuke, line 185:
```suggestion
    previous_matrix = matrix
```

parastell/utils.py
```diff
@@ -101,3 +129,70 @@ def filter_kwargs(
...
+
+    for step in range(steps):
+        smoothed_matrix = np.minimum(
+            previous_iteration,
+            gaussian_filter(
+                previous_iteration,
```
gonuke, line 191:
```suggestion
                previous_matrix,
```

parastell/utils.py
```diff
@@ -101,3 +129,70 @@ def filter_kwargs(
...
+                sigma=sigma,
+                mode="wrap",
+            ),
+        )
+        previous_iteration = smoothed_matrix
```
gonuke, line 196:
```suggestion
        previous_matrix = smoothed_matrix
```

gonuke (changes requested):
A little final polish...

parastell/utils.py
connoramoreno: Something similar to the above snippet can be applied to the poloidal profile at the beginning of the period as well.

parastell/magnet_coils.py
```diff
@@ -436,6 +409,83 @@ def create_magnet(self):
...
+        min_tor_ang = np.min(toroidal_angles)
+        max_tor_ang = np.max(toroidal_angles)
+
+        # Determine if filament toroidal extent overlaps with that of model
+        if (min_tor_ang >= lower_bound or min_tor_ang <= upper_bound) or (
+            max_tor_ang >= lower_bound or max_tor_ang <= upper_bound
+        ):
+            in_toroidal_extent = True
+        else:
+            in_toroidal_extent = False
```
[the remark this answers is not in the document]
connoramoreno: I tried this before and, unfortunately, because of the way we need to define `lower_bound` and `upper_bound` in `MagnetSet._filter_coils`, we need an `or` operator between comparisons and not an `and` operator (the chain acts as an `and`). It's difficult to explain in a single comment but it has to do with which range toroidal angles are defined on and the edge case that the model's toroidal extent is > 180 degrees. There may be a better way of defining lower and upper bounds to check against that allows comparison chaining, but I can't think of one.

parastell/utils.py
```diff
@@ -1,70 +1,102 @@
...
+    matrix[:, -1] = matrix[:, 0]
 
-    init_ang = ang_list[0]
-    final_ang = ang_list[-1]
-    ang_extent = final_ang - init_ang
+    # Ensure poloidal symmetry at beginning of period
+    matrix[0] = np.concatenate(
+        [
+            matrix[0, : math.ceil(num_columns / 2)],
+            np.flip(matrix[0, : math.floor(num_columns / 2)]),
```
gonuke, line 49:
```suggestion
            # ceil and floor ensure middle element of odd sized array is included only once
            matrix[0, : math.ceil(num_columns / 2)],
            np.flip(matrix[0, : math.floor(num_columns / 2)]),
```

gonuke (changes requested):
One more comment for posterity

gonuke (approved):
Thanks for these updates @connoramoreno

PR #154, opened 2024-09-13
Prevent pushes from modifying svalinn's parastell docker image

bquan0:
Previously, the location to which the docker image was being pushed was hardcoded to the `svalinn` user. I have changed all instances of `svalinn` in `docker_publish.yml` to `${{ github.repository_owner }}` so that the docker images being built on forks will be pushed to the user's fork, and will not overwrite svalinn's docker image. 

This will allow contributors to test changes without worrying about the changes breaking svalinn's docker image. Also, once a contributor's PR is merged, `docker_publish.yml` will run on the svalinn user's repo and it will update the svalinn user's docker image. 

To see this working, see this `parastell-ci` [image](https://github.com/bquan0/parastell/pkgs/container/parastell-ci) that I now have as part of my packages.

PR #159, opened 2024-09-19
Add tags to docker images to prevent branches from overwriting Parastell docker images

bquan0:
Currently, `docker_publish.yml` pulls and pushes docker images to a hardcoded image name and tag, which is the official parastell docker image. This can be problematic if a branch changes something and pushes a docker image that doesn't work.

I have changed `docker_publish.yml` to now pull and push docker images to the same image name, but with a tag based on the branch name if the branch is not `main`. This way, contributors can make changes without worrying about breaking the official image. 

This PR is a continuation of #154, with the difference being that this PR was pushed to a branch in the repo owned by `svalinn` instead of a branch in my forked repo.

### DIFF
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -39,8 +39,8 @@ jobs:
       - name: Add extra tag if NOT on the main branch
         if: github.ref_name != 'main'
         run: |
-          echo "EXTRA_TAG=-$github.ref_name" >> $GITHUB_ENV
-          echo "EXTRA_TAG_CI=:$github.ref_name" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


bquan0:
The workflows have passed, and there are now new `docker-build-tweak` tags for the [parastell image](https://github.com/svalinn/parastell/pkgs/container/parastell) and the [parastell-ci image](https://github.com/svalinn/parastell/pkgs/container/parastell-ci) (since that's the name of this branch). The `Test CI Image` [workflow run](https://github.com/svalinn/parastell/actions/runs/10950295081/job/30405273718?pr=159) also uses the image with the correct tag. 

Let me know if I should change the tag name generation.

gonuke (approved):
Thanks @bquan0

PR #160, opened 2024-09-26
import ivb components first so vol ids start at 1

Edgar-21:
It is convenient to have the cell ids start at 1 with the plasma chamber, makes it more obvious which cells will be which in openmc


connoramoreno (approved):
Massive change. LGTM, thanks @Edgar-21!

PR #162, opened 2024-09-27
Poloidal degree change

FusionSandwich:
@connoramoreno please take a look and let us know what you think.


parastell/nwl_utils.py
```diff
@@ -186,6 +186,11 @@ def flux_coords(plas_eq, wall_s, coords, num_threads):
             for theta_coord_chunk in theta_coord_chunks
             for theta_coord in theta_coord_chunk
         ]
+    # Ensures theta_coords are all positive (add 360 degrees where needed)
+    theta_coords = [
+        (theta if theta >= 0 else theta + 2 * np.pi) for theta in theta_coords
+    ]
```
connoramoreno, line 192:
```suggestion
    theta_coords = (theta_coords + 2 * np.pi) % 2*np.pi
```
Avoids the use of conditionals

connoramoreno (changes requested):
Just a small suggestion for the computation of positive angles; otherwise LGTM!

parastell/nwl_utils.py
connoramoreno: You should be able to directly commit this suggestion

FusionSandwich: That's pretty slick!

connoramoreno:
I can't merge until this is formatted with black

connoramoreno:
My suggestion must've done something to the formatting

parastell/nwl_utils.py
```diff
@@ -186,6 +186,9 @@ def flux_coords(plas_eq, wall_s, coords, num_threads):
             for theta_coord_chunk in theta_coord_chunks
             for theta_coord in theta_coord_chunk
         ]
+    # Ensures theta_coords are all positive (add 360 degrees where needed)
+    theta_coords = (theta_coords + 2 * np.pi) % 2*np.pi
```
gonuke, line 190: will they need to be reordered/sorted?

Edgar-21: Doesn't need to be reordered, theta_coords is just a list of the same length as the list of surface crossings from the surface source that stores the poloidal angle for each surface crossing, all we are doing here is changing the notation of theta, each theta_coord still corresponds to the same surface crossing.

For reference, here is what the NWL looks like from before:
image
And after:
image

The first has been rescaled as noted in its title, but looking at the overall profile the coordinate shift can be seen

gonuke, line 190: I think this is why black is complaining.
```suggestion
    theta_coords = (theta_coords + 2 * np.pi) % 2 * np.pi
```

connoramoreno: Thanks!

connoramoreno (approved):
Looks good, thanks @FusionSandwich

PR #164, opened 2024-10-08
Use list comprehension for shift the theta coordinate from 0 to 2pi

Edgar-21:
I experienced an error with nwl_utils.py as currently written, since theta_coords is a list. This approach, or changing theta_coords to a numpy array will alleviate that error.

Fixed a missing comma in invessel_build.py

### DIFF
--- a/parastell/nwl_utils.py
+++ b/parastell/nwl_utils.py
@@ -187,7 +187,9 @@ def flux_coords(plas_eq, wall_s, coords, num_threads):
             for theta_coord in theta_coord_chunk
         ]
     # Ensures theta_coords are all positive (add 360 degrees where needed)
-    theta_coords = (theta_coords + 2 * np.pi) % 2 * np.pi
+    theta_coords = [
+        (theta_coord + 2 * np.pi) % 2 * np.pi for theta_coord in theta_coords
+    ]
 
     return phi_coords.tolist(), theta_coords
 
--- a/parastell/nwl_utils.py
+++ b/parastell/nwl_utils.py
@@ -145,7 +145,7 @@ def find_coords(data):
             bounds=[(-np.pi, np.pi)],
             args=(vmec, wall_s, coords[0], coords[1]),
         )
-        thetas.append(theta.x[0])
+        thetas.append((theta.x[0] + 2 * np.pi) % (2 * np.pi))
     return thetas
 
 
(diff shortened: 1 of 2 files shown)


parastell/nwl_utils.py
```diff
@@ -187,7 +187,9 @@ def flux_coords(plas_eq, wall_s, coords, num_threads):
             for theta_coord in theta_coord_chunk
         ]
     # Ensures theta_coords are all positive (add 360 degrees where needed)
-    theta_coords = (theta_coords + 2 * np.pi) % 2 * np.pi
+    theta_coords = [
+        (theta_coord + 2 * np.pi) % 2 * np.pi for theta_coord in theta_coords
+    ]
```
connoramoreno, line 192: Might be better to change `theta_coords` to a NumPy array

connoramoreno (commented):
Might be better practice to use a NumPy array instead of relying on list comprehension, as you mentioned.

parastell/nwl_utils.py
Edgar-21: Moving to arrays rather than lists would be somewhat large rework, as we make use of append to deal with the variable length iterables that come from the multithreading approach, I'd prefer to keep this PR as a minimal change to get things into a working state.

Regardless, I realized we can just put this coordinate shift in the function where we calculate theta, which I think is a bit more elegant.

parastell/nwl_utils.py
```diff
@@ -145,7 +145,7 @@ def find_coords(data):
             bounds=[(-np.pi, np.pi)],
             args=(vmec, wall_s, coords[0], coords[1]),
         )
-        thetas.append(theta.x[0])
+        thetas.append((theta.x[0] + 2 * np.pi) % (2 * np.pi))
```
connoramoreno, line 148: For our future benefit, it's best to add a comment here explaining (briefly) this conversion

parastell/nwl_utils.py
Edgar-21: The whole nwl_utils could possibly be reworked similar to what Connor has done for radial_distance_utils, but I think that would be better for a different PR.

connoramoreno (changes requested):
I'm happy with this change if @gonuke approves. See my comment about documentation.

connoramoreno (approved):
LGTM! I'll wait to see what Paul has to think before merging. Thanks @Edgar-21

gonuke (approved):

PR #165, opened 2024-10-09
Modify and update magnet meshing workflow

connoramoreno:
Makes the following changes in an effort to improve the magnet meshing workflow, as well as fix some oversights in a previous PR:

- Changes magnet meshing algorithm in Cubit to one that gives greater control over mesh parameters, adding additional meshing arguments
- Puts checks in magnet meshing workflow to determine whether magnets have already been imported into Cubit, and if not, to properly set volume IDs
- Modifies magnet meshing call in Stellarator class to incorporate additional meshing arguments
- Moves magnet and IVB STEP imports in Cubit to corresponding objects
- Updates magnet STEP import in Cubit to properly set volume IDs, reflecting recent changes in STEP import order
- Updates tests to check that volume IDs are properly set

### DIFF
--- a/parastell/cubit_io.py
+++ b/parastell/cubit_io.py
@@ -97,6 +97,10 @@ def export_mesh_cubit(filename, export_dir=""):
     subprocess.run(f"mbconvert {exo_path} {h5m_path}", shell=True)
     Path.unlink(exo_path)
 
+    # Delete any meshes present to prevent inclusion in future Cubit mesh
+    # exports
+    cubit.cmd(f"delete mesh volume all propagate")
+
 
 def export_dagmc_cubit_legacy(
     faceting_tolerance=None,
@@ -177,3 +181,7 @@ def export_dagmc_cubit_native(
 
     export_path = Path(export_dir) / Path(filename).with_suffix(".h5m")
     cubit.cmd(f'export cf_dagmc "{export_path}" overwrite')
+
+    # Delete any meshes present to prevent inclusion in future Cubit mesh
+    # exports
+    cubit.cmd(f"delete mesh volume all propagate")
--- a/parastell/cubit_io.py
+++ b/parastell/cubit_io.py
@@ -79,7 +79,7 @@ def export_cub5(filename, export_dir=""):
     cubit.cmd(f'save cub5 "{export_path}" overwrite')
 
 
-def export_mesh_cubit(filename, export_dir=""):
+def export_mesh_cubit(filename, export_dir="", delete_upon_export=True):
     """Exports Cubit mesh to H5M file format, first exporting to Exodus format
     via Coreform Cubit and converting to H5M via MOAB.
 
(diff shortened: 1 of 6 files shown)


parastell/magnet_coils.py
```diff
@@ -291,6 +313,10 @@ def export_mesh(self, mesh_filename="magnet_mesh", export_dir=""):
             filename=mesh_filename, export_dir=export_dir
         )
 
+        # Delete magnet mesh to prevent inclusion in future Cubit mesh exports
+        volume_ids_str = " ".join(str(id) for id in self.volume_ids)
+        cubit.cmd(f"delete mesh volume {volume_ids_str} propagate")
```
connoramoreno, line 318: While not useful now, it will be when we incorporate in-vessel component meshing (coming up soon)

parastell/cubit_io.py
```diff
@@ -97,6 +97,10 @@ def export_mesh_cubit(filename, export_dir=""):
     subprocess.run(f"mbconvert {exo_path} {h5m_path}", shell=True)
     Path.unlink(exo_path)
 
+    # Delete any meshes present to prevent inclusion in future Cubit mesh
+    # exports
+    cubit.cmd(f"delete mesh volume all propagate")
```
gonuke, line 102: Do we always want to do this? Or should it be an option for the function?

parastell/cubit_io.py
```diff
@@ -177,3 +181,7 @@ def export_dagmc_cubit_native(
 
     export_path = Path(export_dir) / Path(filename).with_suffix(".h5m")
     cubit.cmd(f'export cf_dagmc "{export_path}" overwrite')
+
+    # Delete any meshes present to prevent inclusion in future Cubit mesh
+    # exports
+    cubit.cmd(f"delete mesh volume all propagate")
```
gonuke, line 187: As above, do we always want to do this?

parastell/magnet_coils.py
```diff
@@ -238,6 +238,19 @@ def build_magnet_coils(self):
 
         self._cut_magnets()
 
+    def import_step_cubit(self):
+        """Import STEP file for magnet set into Coreform Cubit."""
+        if cubit_io.initialized:
+            first_vol_id = cubit.get_last_id("volume") + 1
+        else:
+            first_vol_id = 1
```
gonuke, line 246:
```suggestion
        first_vol_id = 1
        if cubit_io.initialized:
            first_vol_id += cubit.get_last_id("volume")
```

parastell/magnet_coils.py
```diff
@@ -238,6 +238,19 @@ def build_magnet_coils(self):
...
+
+        last_vol_id = cubit_io.import_step_cubit(
+            self.step_filename, self.export_dir
+        )
+
+        self.volume_ids = range(first_vol_id, last_vol_id + 1)
```
gonuke, line 252: Should this be a list?
```suggestion
        self.volume_ids = [range(first_vol_id, last_vol_id + 1)]
```

gonuke (changes requested):
a few little comments

parastell/cubit_io.py
connoramoreno: Supporting an option is probably a better idea

gonuke (approved):
Thanks @connoramoreno - LGTM

PR #167, opened 2024-10-10
Fix MOAB canonical ordering during element splitting

### LINKED ISSUE #166: Negative volumes in source meshes
Source mesh elements have been observed to sometimes have negative volume, particularly near the magnetic axis where wedge splitting occurs. Investigation into the source_mesh.py code has revealed that the implementation of MOAB canonical ordering is incorrect for wedges.

connoramoreno:
Corrects implementation of MOAB canonical ordering. Changes are as follows:

- Changes MOAB canonical ordering for both wedges and hexahedra to coincide with element splitting shown in ParaStell publication (Moreno, Bader, Wilson, 2024)
- Adds documentation explaining MOAB canonical ordering for posterity

Closes #166

### DIFF
--- a/tests/test_source_mesh.py
+++ b/tests/test_source_mesh.py
@@ -22,7 +22,9 @@ def source_mesh():
 
     vmec_obj = read_vmec.VMECData(vmec_file)
 
-    mesh_size = (4, 8, 4)
+    # Set mesh size to minimum that maintains element aspect ratios that do not
+    # result in negative volumes
+    mesh_size = (3, 31, 31)
     toroidal_extent = 90.0
 
     source_mesh_obj = sm.SourceMesh(vmec_obj, mesh_size, toroidal_extent)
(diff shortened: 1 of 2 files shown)


connoramoreno:
I'm not sure how we'd test canonical ordering of tetrahedra directly but one test that we should probably add is to ensure no elements with negative volume are present

tests/test_source_mesh.py
```diff
@@ -68,6 +70,31 @@ def test_vertices(source_mesh):
...
+    ) + tets_per_hex * (num_s - 2) * (num_theta - 1) * (num_phi - 1)
+    num_neg_vols_exp = 0
+
+    remove_files()
+
+    source_mesh.create_vertices()
+    source_mesh.create_mesh()
+
+    assert len(source_mesh.volumes) == num_elements_exp
+    assert len([i for i in source_mesh.volumes if i < 0]) == num_neg_vols_exp
```
gonuke, line 93: Nice - does this fail with the old/wrong canonical ordering?

gonuke (commented):
A quick question before we move forward....

tests/test_source_mesh.py
connoramoreno: With the old canonical ordering, this doesn't fail on the test equilibrium, but it does fail with WISTELL-D. On WISTELL-D using these parameters, the number of elements is as expected but 108 elements (1.47%) have negative volume.

gonuke (approved):
Thanks @connoramoreno

PR #171, opened 2024-10-25
Use spline surface fitting for in-vessel component CAD generation

### LINKED ISSUE #26: Allow builds to begin at toroidal angle other than 0
Right now, ParaStell is set up such that all builds start at a toroidal angle of 0. To make ParaStell more general, the toroidal extent of the model should be supplied as a tuple, with initial angle allowed to be any value (positive or negative).

### LINKED ISSUE #170: In-vessel build CAD generation issue when `toroidal_angles` exceeds 180 degrees
In-vessel component CAD generation can be erroneous when the toroidal extent of the build exceeds 180 degrees. While this has historically been solved by using the `repeat` parameter, @tokamaster pointed out that this workflow cannot work for single-period stellarators. It would be useful to modify the CAD generation routine to successfully build 360-degree geometries without using the `repeat` parameter.

connoramoreno:
### Overview

Modifies the in-vessel component CAD generation workflow to use CadQuery’s spline surface fitting routine, instead of its lofting routine, to define the outer surfaces of each in-vessel component (IVC). The benefits of this change are several-fold:

- Facilitates radial build definitions that begin at toroidal and poloidal angles other than 0 degrees
- Builds can span any toroidal extent – no longer need to repeat periods
- The spline surface fitting routine tends to be more robust against steep gradients in IVC thickness matrices
- IVC outer surfaces tend to be smoother
- IVC file sizes are reduced

### List of changes in branch:

- Modifies ParaStell’s `Surface.generate_surface` method to use CadQuery’s `Face.makeSplineApprox` surface spline technique instead of its `Solid.makeLoft` routine
- Modifies ParaStell’s `InVesselBuild.generate_components` method to construct CAD solids by assembling necessary surfaces, instead of performing a cut operation
- Adds `InVesselBuild._check_edge` method
- Removes the now extraneous `Rib` class and moves any necessary methods to the `Surface` class
- Renames `num_ribs` and `num_rib_pts` parameters due to removal of `Rib` class
- Removes the now extraneous `repeat` parameter
- Introduces new parameter `transpose_fit`, which is a flag to indicate that the 2-D iterable of points through which each spline surface is fit should be transposed. This swaps the order in which the grid axes are iterated over. For a yet-to-be-understood reason, this can fix errant CAD generation.
- Adds `min_mesh_size` and `max_mesh_size` arguments to the `InVesselBuild.export_cad_to_dagmc` method
- Modifies the `utils.normalize` function to allow 3-D NumPy arrays
- Updates documentation

### Issues to be solved:

- While CAD generation more robust than lofting workflow, still temperamental to some degree
- In-vessel build test failing for unusual reason

Closes #26
Closes #170

### Illustration of new CAD generation workflow:

Screenshot 2024-10-25 at 11 53 47 AM

Figure 1. Create separate surfaces for outer surface and end caps (plasma/chamber).

Screenshot 2024-10-25 at 11 54 01 AM

Figure 2. Combine surfaces into single solid (plasma/chamber).

Screenshot 2024-10-25 at 11 54 32 AM

Figure 3. Create separate surfaces for inner surface (outer surface of interior component), outer surface, and end caps.

Screenshot 2024-10-25 at 11 54 46 AM

Figure 4. Combine surfaces into single solid.

### Demonstration of new capabilities:

Screenshot 2024-10-25 at 12 34 27 PM

Figure 5. 360-degree geometry.

Screenshot 2024-10-25 at 12 40 15 PM

Figure 6. 360-degree geometry (cut at mid-plane, visualized in ParaView).

Screenshot 2024-10-25 at 11 42 07 AM

Figure 7. In-vessel build with thickness matrix of component defined as `200 * numpy.random.rand(9, 9)`.

### DIFF
--- a/parastell/invessel_build.py
+++ b/parastell/invessel_build.py
@@ -1,6 +1,7 @@
 import argparse
 from pathlib import Path
 
+import math
 import numpy as np
 from scipy.interpolate import RegularGridInterpolator
 
@@ -75,16 +76,19 @@ class InVesselBuild(object):
             logger is supplied, a default logger will be instantiated.
 
     Optional attributes:
-        repeat (int): number of times to repeat build segment for full model
-            (defaults to 0).
-        num_ribs (int): total number of ribs over which to loft for each build
-            segment (defaults to 61). Ribs are set at toroidal angles
-            interpolated between those specified in 'toroidal_angles' if this
-            value is greater than the number of entries in 'toroidal_angles'.
-        num_rib_pts (int): total number of points defining each rib spline
-            (defaults to 67). Points are set at poloidal angles interpolated
-            between those specified in 'poloidal_angles' if this value is
-            greater than the number of entries in 'poloidal_angles'.
+        toroidal_grid_size (int): number of toroidal angle grid points defining
+            point clouds over which spline surfaces are fit. If this value is
+            greater than the length of "toroidal_angles", additional grid
+            points are set at toroidal angles interpolated between those
+            specified in "toroidal_angles".
+        poloidal_grid_size (int): number of poloidal angle grid points defining
+            point clouds over which spline surfaces are fit. If this value is
+            greater than the length of "poloidal_angles", additional grid
+            points are set at poloidal angles interpolated between those
+            specified in "poloidal_angles".
+        transpose_fit (bool): flag to indicate whether the 2-D iterable of
+            points through which each component's spline surface is fit should
+            be transposed. Can sometimes fix errant CAD generation.
         scale (float): a scaling factor between the units of VMEC and [cm]
             (defaults to m2cm = 100).
     """
@@ -95,15 +99,15 @@ class InVesselBuild(object):
         self.vmec_obj = vmec_obj
         self.radial_build = radial_build
 
-        self.repeat = 0
-        self.num_ribs = 61
-        self.num_rib_pts = 67
+        self.toroidal_grid_size = 61
+        self.poloidal_grid_size = 61
+        self.transpose_fit = False
         self.scale = m2cm
 
         for name in kwargs.keys() & (
-            "repeat",
-            "num_ribs",
-            "num_rib_pts",
+            "toroidal_grid_size",
+            "poloidal_grid_size",
+            "transpose_fit",
             "scale",
         ):
             self.__setattr__(name, kwargs[name])
@@ -126,22 +130,6 @@ class InVesselBuild(object):
     @logger.setter
     def logger(self, logger_object):
         self._logger = log.check_init(logger_object)
-
-    @property
-    def repeat(self):
-        return self._repeat
-
-    @repeat.setter
-    def repeat(self, num):
-        self._repeat = num
-        if (self._repeat + 1) * self.radial_build.toroidal_angles[-1] > 360.0:
-            e = AssertionError(
-                "Total toroidal extent requested with repeated geometry "
-                'exceeds 360 degrees. Please examine the "repeat" parameter '
-                'and the "toroidal_angles" parameter of "radial_build".'
-            )
-            self._logger.error(e.args[0])
-            raise e
 
     def _interpolate_offset_matrix(self, offset_mat):
         """Interpolates total offset for expanded angle lists using cubic spline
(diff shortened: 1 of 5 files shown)


Edgar-21:
I'd just like to mention that I started using this this weekend and have found it to be very useful, thanks @connoramoreno

parastell/invessel_build.py
```diff
@@ -226,39 +217,71 @@ def calculate_loci(self):
...
+        Returns:
+            (bool): flag to indicate whether edge is located at either end of
+                stellarator build.
+        """
+        x = edge.Center().x
+        y = edge.Center().y
+
+        toroidal_angle = np.arctan2(y, x)
+        """if self.radial_build.toroidal_angles[0] >= 0.0:
+            toroidal_angle = (toroidal_angle + 2 * np.pi) % (2 * np.pi)"""
```
gonuke, line 237: Do we need this?  Seems like we might want it, in general?

parastell/invessel_build.py
```diff
@@ -226,39 +217,71 @@ def calculate_loci(self):
 
         [surface.calculate_loci() for surface in self.Surfaces.values()]
 
+    def _check_edge(self, edge):
```
gonuke, line 220: Functions returning booleans may be better with a name that indicates the outcome, like `_is_near_end`

parastell/invessel_build.py
```diff
@@ -226,39 +217,71 @@ def calculate_loci(self):
...
+        inner_surface = None
 
         for name, surface in self.Surfaces.items():
             outer_surface = surface.generate_surface()
+            edge_list = [
+                edge
+                for edge in outer_surface.Edges()
+                if self._check_edge(edge)
+            ]
+            wire_list = [cq.Wire.assembleEdges([edge]) for edge in edge_list]
```
gonuke, line 265: Can't we just combine these? I don't see `edge_list` used elsewhere:
```suggestion
            wire_list = [cq.Wire.assembleEdges([edge]) for edge in outer_surface.Edges()
                if self._check_edge(edge)]
```

parastell/invessel_build.py
```diff
@@ -460,8 +446,11 @@ def _vmec2xyz(self, poloidal_offset=0):
         """
         return self.scale * np.array(
             [
-                self.vmec_obj.vmec2xyz(self.s, theta, self.phi)
-                for theta in (self.theta_list + poloidal_offset)
+                [
```
gonuke, line 449: Need to update the docstring to match this new behavior

parastell/invessel_build.py
```diff
@@ -619,14 +620,15 @@ def poloidal_angles(self, angle_list):
             self._logger.error(e.args[0])
             raise e
 
-        self._poloidal_angles = angle_list
-        if self._poloidal_angles[-1] - self._poloidal_angles[0] > 360.0:
+        if angle_list[-1] - angle_list[0] != 360.0:
```
gonuke, line 623: Exactly equal in floating point math is tricky.

Should we check that these are close and then move one to make them exact if they are close, or make an error if they aren't close?

gonuke (changes requested):
A few minor changes, I think.  Do we need to update examples, as well?

Will the change in interface break all the old parastell usage?

parastell/invessel_build.py
connoramoreno: I think I meant to delete that commented code block. In my local version of this branch (yet to be pushed) I created a utility function to check whether an input value is within a given range, accounting for a periodic domain, removing the need to ensure that angles are positive.

parastell/invessel_build.py
connoramoreno: I'll be rolling this branch back a bit to preserve our lofting functionality, instead adding this spline surface capability as an additional feature. That will involve restoring the `Rib` class and this method will maintain its old behavior.

parastell/invessel_build.py
connoramoreno: Good point. I think as long as they're close (within a reasonably small tolerance) it's fine. We should raise the error if they're not close.

connoramoreno:
Closing since there are better alternative modeling schemes

PR #172, opened 2024-11-01
Implement alternating division schemes between adjacent hexahedra and wedges

### LINKED ISSUE #168: Source mesh total source strength scaling incorrectly with mesh discretization
The expected scaling behavior for total source strength (sum of `SourceMesh.strengths`) with increasing mesh discretization is asymptotic. One would expect that with finer discretization, total source strength would approach some singular value (hopefully, the true value).

In particular, what I would expect (though please check my logic) is that closed flux surface (CFS) discretization would have the largest impact on source strength accuracy, since source strength is directly dependent on CFS value. I would then expect poloidal and toroidal angle discretizations to have a lesser though still asymptotic impact since those are mostly related to accurately representing the shape of the plasma.

However, in the current source mesh implementation, as @Edgar-21 pointed out, this is not the observed behavior. Varying each mesh dimension independently, I observed the trends in the following plots, with fusion power representing total source strength.

Screenshot 2024-10-10 at 2 16 56 PM

Figure 1. Calculated fusion power as a function of CFS discretization. Number of poloidal grid points and number of toroidal grid points are kept constant at 61 each.

Screenshot 2024-10-10 at 2 17 02 PM

Figure 2. Calculated fusion power as a function of poloidal angle discretization. Number of CFS grid points and number of toroidal grid points are kept constant at 11 and 61, respectively.

Screenshot 2024-10-10 at 2 17 12 PM

Figure 3. Calculated fusion power as a function of toroidal angle discretization. Number of CFS grid points and number of poloidal grid points are kept constant at 11 and 61, respectively.

Doing the same for the source mesh implementation in PR #167, we observe an entirely different set of trends. This may warrant some discussion.

Screenshot 2024-10-10 at 2 17 20 PM

Figure 4. Calculated fusion power as a function of CFS discretization. Number of poloidal grid points and number of toroidal grid points are kept constant at 61 each.

Screenshot 2024-10-10 at 2 17 24 PM

Figure 5. Calculated fusion power as a function of poloidal angle discretization. Number of CFS grid points and number of toroidal grid points are kept constant at 11 and 61, respectively.

Screenshot 2024-10-10 at 2 17 27 PM

Figure 6. Calculated fusion power as a function of toroidal angle discretization. Number of CFS grid points and number of poloidal grid points are kept constant at 11 and 61, respectively.

connoramoreno:
Modifies the source mesh workflow to alternate the scheme by which hexahedra and wedges are split into tetrahedra. The alternating schemes ensure that the faces of adjacent tetrahedra, but not belonging to the same hexahedron or wedge, match one another. This avoids gaps and overlaps between adjacent non-planar hexahedron and wedge faces. This alignment is ensured for adjacent wedge-wedge, wedge-hexahedron, and hexahedron-hexahedron pairs. Code changes are as follows:

- Introduces `alternate_scheme`, `toroidal_alt_scheme`, and `cfs_alt_scheme` boolean parameters to control whether an alternate division scheme should be used for a given element
- Introduces an alternate division scheme, in the form of MOAB canonical indices defining the tetrahedral division of a given element, for both hexahedra and wedges
- Introduces checks to ensure the number of poloidal grid intervals and (conditionally) the number of toroidal grid intervals are even
- Modifies naming convention of flux-coordinate indices, lists, etc. to be more intuitive
- Modifies vertex indexing scheme to be more intuitive
- Reduces toroidal extent of mesh in `test_source_mesh` unit tests

The alternate splitting schemes are illustrated below. Plots will follow to demonstrate expected behavior of total volume and total source strength as a function of mesh discretization.

(a)Screenshot 2024-11-01 at 2 59 13 PM
(b)Screenshot 2024-11-01 at 3 07 21 PM

Figure 1. Alternate schemes for adjacent hexahedra. (a) Original and (b) alternate.

(a)Screenshot 2024-11-01 at 3 24 55 PM
(b)Screenshot 2024-11-01 at 3 22 00 PM

Figure 2. Alternate schemes for adjacent wedges. (a) Original and (b) alternate.

Closes #168

### DIFF
--- a/parastell/source_mesh.py
+++ b/parastell/source_mesh.py
@@ -85,10 +85,10 @@ class SourceMesh(object):
             any closed flux surface label, s, poloidal angle, theta, and
             toroidal angle, phi.
         mesh_size (tuple of int): number of grid points along each axis of
-            flux-coordinate space, in the order (num_s, num_theta, num_phi).
-            'num_s' is the number of closed flux surfaces for vertex locations
-            in each toroidal plane. 'num_theta' is the number of poloidal
-            angles for vertex locations in each toroidal plane. 'num_phi' is
+            flux-coordinate space, in the order (num_cfs_pts, num_poloidal_pts, num_toroidal_pts).
+            'num_cfs_pts' is the number of closed flux surfaces for vertex locations
+            in each toroidal plane. 'num_poloidal_pts' is the number of poloidal
+            angles for vertex locations in each toroidal plane. 'num_toroidal_pts' is
             the number of toroidal angles for planes of vertices.
         toroidal_extent (float): extent of source mesh in toroidal direction
             [deg].
@@ -112,9 +112,9 @@ class SourceMesh(object):
 
         self.logger = logger
         self.vmec_obj = vmec_obj
-        self.num_s = mesh_size[0]
-        self.num_theta = mesh_size[1]
-        self.num_phi = mesh_size[2]
+        self.num_cfs_pts = mesh_size[0]
+        self.num_poloidal_pts = mesh_size[1]
+        self.num_toroidal_pts = mesh_size[2]
         self.toroidal_extent = toroidal_extent
 
         self.scale = m2cm
@@ -134,11 +134,11 @@ class SourceMesh(object):
         self._create_mbc()
 
     @property
-    def num_theta(self):
-        return self._num_theta
+    def num_poloidal_pts(self):
+        return self._num_poloidal_pts
 
-    @num_theta.setter
-    def num_theta(self, value):
+    @num_poloidal_pts.setter
+    def num_poloidal_pts(self, value):
         if value % 2 != 1:
             e = AttributeError(
                 "To ensure that tetrahedral faces are coincident at the end of "
@@ -148,15 +148,15 @@ class SourceMesh(object):
             )
             self._logger.error(e.args[0])
             raise e
-        self._num_theta = value
+        self._num_poloidal_pts = value
 
     @property
-    def num_phi(self):
-        return self._num_phi
+    def num_toroidal_pts(self):
+        return self._num_toroidal_pts
 
-    @num_phi.setter
-    def num_phi(self, value):
-        self._num_phi = value
+    @num_toroidal_pts.setter
+    def num_toroidal_pts(self, value):
+        self._num_toroidal_pts = value
 
     @property
     def toroidal_extent(self):
@@ -169,7 +169,7 @@ class SourceMesh(object):
             self._logger.error(e.args[0])
             raise e
 
-        if angle == 360.0 and self._num_phi % 2 != 1:
+        if angle == 360.0 and self._num_toroidal_pts % 2 != 1:
             e = AttributeError(
                 "To ensure that tetrahedral faces are coincident at the end of "
                 "the closed toroidal loop, the number of toroidal intervals "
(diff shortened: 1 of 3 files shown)


parastell/source_mesh.py
```diff
@@ -437,14 +458,18 @@ def create_mesh(self):
         self.mbc.add_entity(self.mesh_set, self.verts)
 
         for phi_idx in range(self.num_phi - 1):
+            # Set alternation flag to true at beginning of each toroidal block
+            self.alt_flag = True
```
gonuke, line 462: Let's define this as a local variable and pass into the methods.

gonuke: a different name maybe useful.... Some algorithms use the nomenclature "red/black"

gonuke (changes requested):
Thanks for the quick fix @connoramoreno

connoramoreno:
Below are plots demonstrating that total source strength (in the form of fusion power, calculated for four periods even though only one period of the source mesh is modeled) and total mesh volume scale as expected with mesh discretization.

Expected behavior:
- Total source strength converges to some value with increasing CFS discretization
- Total volume remains constant with changing CFS discretization
- Total source strength changes reflect changes in total volume with changing poloidal discretization
- Total volume converges to some value with increasing poloidal discretization
- Total source strength changes reflect changes in total volume with changing toroidal discretization
- Total volume converges to some value with increasing toroidal discretization

(a) Screenshot 2024-11-03 at 9 08 51 AM
(b) Screenshot 2024-11-03 at 9 09 15 AM
Figure 1. Behavior of total source strength and volume as a function of CFS discretization. Numbers of poloidal and toroidal grid points are kept constant at 121 each. (a) Total source strength. (b) Total mesh volume.

(a) Screenshot 2024-11-03 at 9 09 05 AM
(b) Screenshot 2024-11-03 at 9 09 19 AM
Figure 2. Behavior of total source strength and volume as a function of poloidal discretization. Numbers of CFS and toroidal grid points are kept constant at 16 and 121, respectively. (a) Total source strength. (b) Total mesh volume.

(a) Screenshot 2024-11-03 at 9 09 10 AM
(b) Screenshot 2024-11-03 at 9 09 24 AM
Figure 3. Behavior of total source strength and volume as a function of toroidal discretization. Numbers of CFS and poloidal grid points are kept constant at 16 and 121, respectively. (a) Total source strength. (b) Total mesh volume.

connoramoreno:
I have another PR lined up once this is merged that will allow users to input their own grids, rather than us enforce uniformly spaced grids. I'd like to include checks on inputs (such as warnings for when grid sizes are not even) in that subsequent PR, since it touches a lot of that part of the code.

gonuke (changes requested):
do we need to keep these different alternative scheme variables? For me it's hard to follow. If we require even numbers of poloidal voxels, I think there may be a cleaner way to do it.

gonuke:
Can we just determine the meshing scheme from whether `s_idx + theta_idx + phi_idx` is even or odd?

gonuke:
Could then have canonical ordering indexed in a list with index 0 and 1 and directly use the one you want indexed by the alternate scheme variable.

parastell/source_mesh.py
```diff
@@ -375,23 +413,38 @@ def _create_tets_from_hex(self, s_idx, theta_idx, phi_idx):
...
+                [idx_list[1], idx_list[3], idx_list[6], idx_list[4]],
+            ],
+            [
+                [idx_list[0], idx_list[2], idx_list[1], idx_list[5]],
+                [idx_list[0], idx_list[3], idx_list[2], idx_list[7]],
+                [idx_list[0], idx_list[7], idx_list[5], idx_list[4]],
+                [idx_list[7], idx_list[2], idx_list[5], idx_list[6]],
+                [idx_list[0], idx_list[2], idx_list[5], idx_list[7]],
+            ],
         ]
```
connoramoreno, line 431: Not sure if this is what you had in mind for indexing the canonical ordering scheme @gonuke?

parastell/source_mesh.py
```diff
@@ -375,23 +413,38 @@ def _create_tets_from_hex(self, s_idx, theta_idx, phi_idx):
...
+        for vertex_ids in canonical_ordering_schemes[scheme_idx]:
             self._create_tet(vertex_ids)
 
     def _create_tets_from_wedge(self, theta_idx, phi_idx):
         """Creates three tetrahedra from defined wedge.
         (Internal function not intended to be called externally)
 
         Arguments:
-            idx_list (list of int): list of wedge vertex indices.
+            theta_idx (int): index defining location along poloidal angle axis.
```
gonuke, line 446: I don't understand the stencil below.  Maybe it's been this way for a while, but I don't see any points in this stencil at a `s_idx=1` which I expect.  
I'm also not sure why this goes from [1,num_phi] istead of [0,num_phi-1]

gonuke (changes requested):
Definitely simpler by making that red/black calculation be calculated in the methods that create the tets, but I have some questions on the choices of indexing.

In particular, I'm not sure (or don't remember) the motivation for `theta_idx` to go from `[1,num_phi]` instead of `[0,num_phi-1]`

parastell/source_mesh.py
connoramoreno: Yes, I agree that it's slightly confusing, and our indexing scheme could use some changes to make it more intuitive. The way we've defined things is that the vertex at the magnetic axis and those on the first closed flux surface all have a `s_idx` of 0. This is also why `theta_idx` begins at 1, to differentiate the first vertex on the first CFS from that on the magnetic axis.

It would make much more sense to have `s_idx = 0` to always refer to the magnetic axis vertices, and `s_idx = 1` refer to the first CFS. As such, we could then have `theta_idx` start from 0. I think I'll also change the naming convention from using to `theta_idx` and `phi_idx` to using `poloidal_idx` and `toroidal_idx`, respectively, to avoid confusion.

parastell/source_mesh.py
```diff
@@ -133,18 +134,54 @@ def __init__(
...
     def toroidal_extent(self, angle):
-        self._toroidal_extent = np.deg2rad(angle)
-        if self._toroidal_extent > 360.0:
-            e = ValueError("Toroidal extent cannot exceed 360.0 degrees.")
+        if angle > 360.0:
+            e = AttributeError("Toroidal extent cannot exceed 360.0 degrees.")
+            self._logger.error(e.args[0])
+            raise e
+
+        if angle == 360.0 and self._num_toroidal_pts % 2 != 1:
```
gonuke, line 173: Later on you compare `_toroidal_extent` to 2 pi.  It might make more sense to convert to radians before this test for consistency.

gonuke (changes requested):
This is all definitely a lot easier to follow!!!

gonuke (approved):
Thanks for this important fix @connoramoreno

PR #176, opened 2024-11-13
export to STL format capability added

davidsosa:
I added the option to export geometry files in STL format for both magnets and in-vessel components.


parastell/invessel_build.py
```diff
@@ -309,6 +309,23 @@ def export_step(self, export_dir=""):
             )
             cq.exporters.export(component, str(export_path))
 
+    def export_stl(self, export_dir=""):
```
gonuke, line 312: Since this function is functionally identical to `export_step` it might be better to rename `export_step` as `export_standard` (or something) and add the export type as an argument.

parastell/magnet_coils.py
```diff
@@ -269,6 +269,29 @@ def export_step(self, step_filename="magnet_set", export_dir=""):
         )
         cq.exporters.export(coil_set, str(export_path))
 
+    def export_stl(self, stl_filename="magnet_set", export_dir=""):
```
gonuke, line 272: Similar comment as above about extending the previous method rather than copy/pasting to a new method.

gonuke (changes requested):
Thanks for this useful addition @davidsosa. I've recommended that we implement it as a single function that exports either STEP or STL since they are identical functions.  There may be an argument to keep both `export_step` and `export_stl`, but have them each call the same function that performs the work with an optional file type.

parastell/invessel_build.py
connoramoreno: You could change `export_step` to `export_cad` and have the function take the file extension as an argument (in string format)

parastell/magnet_coils.py
davidsosa: I see. I did this way because It was usefull for me to get these two format at the same time, but for sure it can be done extending the method instead of duplicating it. I'l have a deeper look on it

gonuke: You could still get both in the same run, by just calling the method twice with different arguments.

gonuke (commented):
This looks good to me. Do we currently test this for STEP? Should we add a test for STL?

parastell/invessel_build.py
```diff
@@ -299,20 +299,24 @@ def import_step_cubit(self):
             vol_id = cubit_io.import_step_cubit(name, self.export_dir)
             data["vol_id"] = vol_id
 
-    def export_step(self, export_dir=""):
-        """Export CAD solids as STEP files via CadQuery.
+    def export_components(self, filetype="step", export_dir=""):
+        """Export CAD solids as STEP or STL files via CadQuery.
```
connoramoreno, line 303: It might be a good idea to include some kind of check on filetype input, unless we want to rely on CadQuery to issue an error if an incorrect filetype is supplied.

parastell/magnet_coils.py
```diff
@@ -250,23 +250,29 @@ def import_step_cubit(self):
 
         self.volume_ids = list(range(first_vol_id, last_vol_id + 1))
 
-    def export_step(self, step_filename="magnet_set", export_dir=""):
-        """Export CAD solids as a STEP file via CadQuery.
+    def export_components(
+        self, filetype="step", filename="magnet_set", export_dir=""
+    ):
```
connoramoreno, line 255: Would it make sense to combine `filename` and `filetype` into a single input argument? This would deviate from our established convention.

parastell/magnet_coils.py
```diff
@@ -250,23 +250,29 @@ def import_step_cubit(self):
 
         self.volume_ids = list(range(first_vol_id, last_vol_id + 1))
 
-    def export_step(self, step_filename="magnet_set", export_dir=""):
-        """Export CAD solids as a STEP file via CadQuery.
+    def export_components(
```
connoramoreno, line 253: Instead of calling this `export_components`, since there's only one component being exported, maybe we should call this something like `export_cad`, and do the same with the in-vessel build for consistency.

connoramoreno (changes requested):
Just some general comments and thoughts on naming and inputs that could possibly use some discussion. We'll also need to update the test suite to reflect the modified API.

PR #177, opened 2024-11-17
Create Facets From Point Cloud, and Volumes From Facets With Cubit

Edgar-21:
Still work in progress, but this is functioning. Uses the point cloud data from surface objects to create facets. These facets are then connected into a surface, and the ends are closed off, forming a volume. Working from outside in, the overlap from volumes is removed, and then adjacent volumes are imprinted and merged. Volume ids are stored in the radial build during building so existing dagmc export functions work no problem. As can be seen from the diff here, this isn't particularly complicated, and it will be nice to have options when the splines are acting up, even if there are perhaps better ways long term to make this faceted representation.

Here's the geometry using the inputs from the parastell example, built using this workflow.
image

### DIFF
--- a/parastell/invessel_build.py
+++ b/parastell/invessel_build.py
@@ -231,6 +231,7 @@ class InVesselBuild(object):
         build by cutting the interior surface solid from the outer surface
         solid for a given component.
         """
+        self.cubit_volumes = False
         self._logger.info(
             "Constructing CadQuery objects for in-vessel components..."
         )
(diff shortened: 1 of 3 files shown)


gonuke:
This looks promising - I wonder if we can do this easier/better/faster directly wit pymoab/MOAB?

parastell/invessel_build.py
```diff
@@ -346,6 +346,82 @@ def export_cad_to_dagmc(self, dagmc_filename="dagmc", export_dir=""):
...
+        print("removing overlap")
+        layers = list(self.radial_build.radial_build.values())
+        for layer, next_layer in zip(
+            reversed(layers[0:-1]), reversed(layers[1:])
+        ):
+            cubit.cmd(
+                "remove overlap volume "
+                f"{layer['vol_id']} {next_layer['vol_id']} modify larger"
+            )
+        cubit.cmd('save cub5 "overlap_removed.cub5" overwrite')
```
connoramoreno, line 412: What overlap is occurring?

parastell/invessel_build.py
```diff
@@ -346,6 +346,82 @@ def export_cad_to_dagmc(self, dagmc_filename="dagmc", export_dir=""):
...
+        # merge layers
+        for layer, next_layer in zip(
+            reversed(layers[0:-1]), reversed(layers[1:])
+        ):
+            print(layer)
+            cubit.cmd(
+                f"imprint volume {layer['vol_id']} {next_layer['vol_id']}"
+            )
+            cubit.cmd(f"merge volume {layer['vol_id']} {next_layer['vol_id']}")
```
connoramoreno, line 421: We should probably make this the way we imprint and merge by default in `Stellarator.export_dagmc`

connoramoreno:
Also, in response to @gonuke's comment, it should be possible to do this in MOAB instead, since we're not really performing any CAD operations (except maybe removing overlaps)

parastell/invessel_build.py
Edgar-21: each volume is made without the through hole, similar to how they are made in CADquery now. This subtracts the inner component from the outer component, working outside in.

Edgar-21:
> This looks promising - I wonder if we can do this easier/better/faster directly wit pymoab/MOAB?
> 
> i.e. without Cubit

I don't know much about pymoab/MOAB, but I'm sure it would be possible. This has the benefit of using tools I already knew and existing code in parastell, plus we still can get CAD solids out of Cubit, so even in the event that we implement a MOAB workflow, it might be nice to retain this.

A MOAB workflow would be faster/more scaleable but would need a lot of additional development work.

FusionSandwich:
> Also, in response to @gonuke's comment, it should be possible to do this in MOAB instead, since we're not really performing any CAD operations (except maybe removing overlaps)

Pymoab version is incoming. It's making meshable geometry, I just need to test the code and clean it up a bit.

connoramoreno:
Should this be closed?

Edgar-21:
No longer relevant with PyDAGMC workflow.

PR #178, opened 2024-11-20
Adds support for meshing in-vessel components

tokamaster:
This pull request introduces a method for generating a tetrahedral mesh of individual in-vessel components and exporting the mesh in .h5m format. The process involves the following steps:

1. Importing the STEP file: The method imports the STEP file corresponding to the component.
2. Mesh Generation: Using Cubit, a tetrahedral mesh is created for the imported component.
3. Exporting Mesh: The resulting meshed volume is exported in .exo format, which is then converted to .h5m using MOAB.

When using the method, the names of the in-vessel components must match the names specified in the radial build dictionary. Also, the mesh size can be customized using the mesh_size argument. More information on mesh sizing functions can be found [here](https://coreform.com/cubit_help/mesh_generation/interval_assignment/automatic_specification.htm).

### DIFF
--- a/parastell/invessel_build.py
+++ b/parastell/invessel_build.py
@@ -346,6 +346,33 @@ class InVesselBuild(object):
 
         model.export_dagmc_h5m_file(filename=str(export_path))
 
+    def export_invessel_component_mesh(
+        self, components, mesh_size=5, import_dir="", export_dir=""
+    ):
+        """Creates a tetrahedral mesh of an in-vessel component volume
+        via Coreform Cubit and exports it as H5M file.
+
+        Arguments:
+            components (array of strings): array containing the name
+                of the in-vessel components to be meshed.
+            mesh_size (int): controls the size of the mesh. Takes values
+                between 1 (finer) and 10 (coarser) (optional, defaults to 5).
+            import_dir (str): directory containing the STEP file of
+                the in-vessel component (optional, defaults to empty string).
+            export_dir (str): directory to which to export the h5m
+                output file (optional, defaults to empty string).
+        """
+        for component in components:
+            vol_id = cubit_io.import_step_cubit(component, import_dir)
+            cubit.cmd(f"volume {vol_id} scheme tetmesh")
+            cubit.cmd(f"volume {vol_id} size auto factor {mesh_size}")
+            cubit.cmd(f"mesh volume {vol_id}")
+            cubit_io.export_mesh_cubit(
+                filename=component,
+                export_dir=export_dir,
+                delete_upon_export=False,
+            )
+
 
 class Surface(object):
     """An object representing a surface formed by lofting across a set of
--- a/parastell/parastell.py
+++ b/parastell/parastell.py
@@ -191,6 +191,27 @@ class Stellarator(object):
                 dagmc_filename=dagmc_filename, export_dir=export_dir
             )
 
+    def export_invessel_component_mesh(
+        self, components, mesh_size=5, import_dir="", export_dir=""
+    ):
+        """Creates a tetrahedral mesh of an in-vessel component volume
+        via Coreform Cubit and exports it as H5M file.
+
+        Arguments:
+            components (array of strings): array containing the name
+                of the in-vessel components to be meshed.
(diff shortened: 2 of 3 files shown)


parastell/cubit_io.py
```diff
@@ -205,3 +205,14 @@ def export_dagmc_cubit_native(
     # exports
     if delete_upon_export:
         cubit.cmd(f"delete mesh volume all propagate")
+
+
+def create_new_cubit_instance():
```
gonuke, line 210: I thin it would be helpful to inser this immediately after `init_cubit()` so that folks can find it and associate it with that method more easily.

gonuke (changes requested):
Thanks @tokamaster - this looks pretty good to me.  I have made one minor request for future readability, but want to make sure @connoramoreno has a chance to chime in as well.

parastell/cubit_io.py
tokamaster: Yeah this makes sense; I'll insert it after init_cubit().

parastell/invessel_build.py
```diff
@@ -346,6 +346,33 @@ def export_cad_to_dagmc(self, dagmc_filename="dagmc", export_dir=""):
 
         model.export_dagmc_h5m_file(filename=str(export_path))
 
+    def export_invessel_component_mesh(
```
connoramoreno, line 349: Since this is part of the `InVesselBuild` class, I don't think it's necessary to specify in the name that this method is for in-vessel components. We may want to shorten the name to something like `export_mesh` for brevity and also to match the method name we use for the magnets.

parastell/parastell.py
```diff
@@ -191,6 +191,27 @@ def export_invessel_build(
                 dagmc_filename=dagmc_filename, export_dir=export_dir
             )
 
+    def export_invessel_component_mesh(
```
connoramoreno, line 194: I think we might want to match the name of this method to that of the class for consistency, i.e., `export_invessel_build_mesh`.

connoramoreno (changes requested):
Thanks for the contribution @tokamaster! I just have a couple small thoughts on naming, otherwise this looks good to me

parastell/invessel_build.py
tokamaster: Yep I agree 'invessel' here is redundant. Yet, I think I would keep 'components' in the name of the method, since it meshes, and exports, specific components and not the whole radial build. Something like `export_component_mesh` would be more appropriate in my opinion. A similar logic follows for `generate_components` method in the `InVesselBuild` class.

parastell/parastell.py
tokamaster: I would still keep 'component' in the method's name since it meshes specific components, defined as an argument, rather than the entire radial build. We could have another method for meshing the entire build, though.

parastell/invessel_build.py
```diff
@@ -346,6 +346,33 @@ def export_cad_to_dagmc(self, dagmc_filename="dagmc", export_dir=""):
...
+        for component in components:
+            vol_id = cubit_io.import_step_cubit(component, import_dir)
+            cubit.cmd(f"volume {vol_id} scheme tetmesh")
+            cubit.cmd(f"volume {vol_id} size auto factor {mesh_size}")
+            cubit.cmd(f"mesh volume {vol_id}")
+            cubit_io.export_mesh_cubit(
+                filename=component,
+                export_dir=export_dir,
+                delete_upon_export=False,
+            )
```
connoramoreno, line 374: I believe we will want `delete_upon_export` to be `True` so that loop iterations do not include meshes from previous iterations in their exports.

connoramoreno (changes requested):
Just one small comment about the call to `export_mesh_cubit` in the `InVesselBuild.export_component_mesh` method.

parastell/invessel_build.py
tokamaster: Ha, good catch, I think this makes sense.

gonuke (approved):
This looks good to me - thanks @tokamaster

PR #180, opened 2024-12-12
Full model DAGMC faceting via CAD-to-DAGMC

connoramoreno:
This branch adds support for a fully open-source, freely available DAGMC faceting workflow via CAD-to-DAGMC. This workflow is implemented as an alternative to the Cubit DAGMC faceting workflow.

Changes included in this branch:

- Introduces a new `Stellarator.export_cad_to_dagmc` method to automate DAGMC model generation via CAD-to-DAGMC and adds it to test and example scripts
- Renames the `Stellarator.export_dagmc` method, which generates DAGMC models via Cubit, to `Stellarator.export_cf_dagmc` and updates syntax in test and example scripts
- Removes the `InVesselBuild.export_cad_to_dagmc` method as it is made redundant by the `Stellarator.export_cad_to_dagmc` method and removes it from test and example scripts
- Modifies the magnet modeling workflow to store CAD solids in the `MagnetSet.coil_solids` parameter for future reference

It should be noted that the CAD-to-DAGMC routine used here enforces that imprinting is skipped, since we have determined based on the way we create geometry that imprinting is unnecessary. [It was found](https://github.com/fusion-energy/cad_to_dagmc/issues/110) that CAD-to-DAGMC's imprinting routine was computationally expensive and required an unreasonable amount of memory for our geometries. Skipping this routine alleviates this computational cost and the meshing is performed in seconds.


connoramoreno:
It should be noted that the "Regular CI testing for parastell / Perform CI Tests" workflow is failing because it is not pulling the Docker image corresponding to this branch

connoramoreno:
@bquan0, do you know how to fix the issue mentioned in the above comment, concerning the proper image tag being used?

bquan0:
It's because of [this line](https://github.com/svalinn/parastell/blob/12cf70afc7e80b0556feb21f32ef552995d47fc1/.github/workflows/ci.yml#L15). `container: ghcr.io/svalinn/parastell-ci` means it always pulls the `latest` image. 
It needs some logic similar to [this line in docker_publish.yml](https://github.com/svalinn/parastell/blob/12cf70afc7e80b0556feb21f32ef552995d47fc1/.github/workflows/docker_publish.yml#L62) where it determines if it needs the extra tag if it's not on the `main` branch.

connoramoreno:
I figured as much! Just unsure how to go about retrieving the proper tag. Would it be done similarly to [this line](https://github.com/svalinn/parastell/blob/12cf70afc7e80b0556feb21f32ef552995d47fc1/.github/workflows/docker_publish.yml#L43)?

bquan0:
I think you would have to run an extra job before `test-dependency-img` that echoes the tag to an `output`, like in [line 46](https://github.com/svalinn/parastell/blob/12cf70afc7e80b0556feb21f32ef552995d47fc1/.github/workflows/docker_publish.yml#L46). This is because ENV variables can't be used in `container:`, I think.

connoramoreno:
Since "**Build & publish Docker image for ParaStell CI / Test CI Image**" runs our test suite when we've made changes to any file affecting our Docker image, I'm going to add a condition to "**Regular CI testing for parastell / Perform CI Tests**" that will prevent its execution (returning the "Success" status) should any of those aforementioned files be modified. In effect, this change will make it such that our CI testing will defer to "**Build & publish Docker image for ParaStell CI / Test CI Image**" to test the code under the condition that the Docker image has been modified.

parastell/magnet_coils.py
```diff
@@ -268,11 +270,30 @@ def export_step(self, step_filename="magnet_set", export_dir=""):
...
+                appended. Note that the material tag will be appended once for
+                each magnet solid.
+
+        Returns:
+            solids (list): updated list including magnet set CadQuery solid
+                objects.
+            mat_tags (list): updated list including magnet set material tags.
+        """
+        solids.extend(self.coil_solids)
+        [mat_tags.append(self.mat_tag) for _ in self.coil_solids]
```
gonuke, line 293: This seems more readable to me
```suggestion
        mat_tags.extend([self.mat_tag] * len(self.coil_solids))
```

parastell/parastell.py
```diff
@@ -14,7 +15,7 @@
 from .utils import read_yaml_config, filter_kwargs, m2cm
 
 build_cubit_model_allowed_kwargs = ["skip_imprint", "legacy_faceting"]
-export_dagmc_allowed_kwargs = [
+export_cf_dagmc_allowed_kwargs = [
```
gonuke, line 18: What does `cf` mean here?

gonuke: I think it's for Coreform?  I wonder if `cubit` would be better than `cf`?

gonuke (changes requested):
Just a couple of little comments.  Are there any important limitations?

parastell/parastell.py
connoramoreno: Yes, `cf` refers to Coreform. I was copying their DAGMC export syntax in older versions of Cubit, but I agree using `cubit` would be clearer in our case.

parastell/parastell.py
```diff
@@ -387,20 +381,21 @@ def export_dagmc(self, filename="dagmc", export_dir="", **kwargs):
...
-                (defaults to 100.0).
+                (optional, defaults to 100.0).
             deviation_angle (float): controls deviation angle of facet from
                 surface (i.e., lesser deviation angle results in more elements
-                in areas with higher curvature) (defaults to 5.0).
+                in areas with higher curvature) (optional, defaults to 5.0).
         """
         cubit_io.init_cubit()
 
         self._logger.info("Exporting DAGMC neutronics model...")
```
gonuke, line 392:
```suggestion
        self._logger.info("Exporting DAGMC neutronics model using Coreform Cubit...")
```

parastell/parastell.py
```diff
@@ -418,6 +413,68 @@ def export_cub5(self, filename="stellarator", export_dir=""):
...
+            filename (str): name of DAGMC output file, excluding '.h5m'
+                extension (optional, defaults to 'dagmc').
+            export_dir (str): directory to which to export DAGMC output file
+                (optional, defaults to empty string).
+            min_mesh_size (float): minimum size of mesh elements (defaults to
+                20).
+            max_mesh_size (float): maximum size of mesh elements (defaults to
+                50).
+        """
+        self._logger.info("Exporting DAGMC neutronics model...")
```
gonuke, line 468:
```suggestion
        self._logger.info("Exporting DAGMC neutronics model with CAD-to-DAGMC ...")
```

parastell/magnet_coils.py
```diff
@@ -268,11 +270,30 @@ def export_step(self, step_filename="magnet_set", export_dir=""):
             self.step_filename
         ).with_suffix(".step")
 
-        coil_set = cq.Compound.makeCompound(
-            [coil.solid for coil in self.magnet_coils]
-        )
+        coil_set = cq.Compound.makeCompound(self.coil_solids)
         cq.exporters.export(coil_set, str(export_path))
 
+    def extract_solids_and_mat_tag(self, solids, mat_tags):
```
gonuke, line 276: In general, I find it a little unsettling to have functions that change the input parameters and pass them back as new output parameters.

I wonder if it's better to just return these solids and tags, and append/extend outside of the function? I guess that can be accomplished by passing in empty lists...??

If we are appending in this function, maybe the function name should be `append_solids_and_mat_tags` rather than `extract_solids_and_mat_tag`?

gonuke (changes requested):
A couple of minor comments to finalize this.

Also do we have any examples using the new capability? I see the examples are all updated, but still only include Cubit cases - is that right?

connoramoreno:
I'll add an example that uses CAD-to-DAGMC

parastell/magnet_coils.py
connoramoreno: Fair point! I think I prefer the former. In the case that we append/extend outside the function, it shouldn't be necessary to provide any input arguments.

parastell/parastell.py
```diff
@@ -418,6 +413,64 @@ def export_cub5(self, filename="stellarator", export_dir=""):
...
+            ivb_solids, ivb_material_names = (
+                self.invessel_build.extract_solids_and_mat_tags()
+            )
+
+        if self.magnet_set:
+            ms_solids, ms_material_names = (
+                self.magnet_set.extract_solids_and_mat_tag()
+            )
+
+        solids = ivb_solids + ms_solids
```
gonuke, line 434: This may fail if there is no IVB or no magnets.  Maybe another test case?

gonuke (changes requested):

gonuke (approved):
This looks great @connoramoreno - thanks for this improvement

PR #190, opened 2025-02-03
Add PyDAGMC model generation

Edgar-21:
This PR adds the ability to generate a DAGMC model of IVB components by building faceted surfaces from the point cloud representation using PyMOAB and PyDAGMC features. The plasma chamber is not currently being generated (i.e. the implicit complement is used) as the problem of closing off the ends for the plasma chamber is not trivial. This could be addressed in a future PR if desired.

To add magnets to this model, a method to merge two separate DAGMC models by re-indexing surface, volume, and group ids is added.

An example script demonstrating the new workflow is added, and tests have been updated.


Edgar-21:
CI failing due to the new PyDAGMC dependency. I don't remember how we get the new dependency in the testing environment.

connoramoreno:
DAGMC gets installed as a sub-dependency but it doesn't seem to be importable for Python. Is there a separate package needed to make DAGMC Python-importable?

connoramoreno:
What we decided was the best course of action is to create a separate PR to introduce the change in dependencies and merge that before the PR for the rest of the code changes (this one)

Edgar-21:
Makes sense, just opened a PR to add PyDAGMC to the environment. Should resolve the CI issue.

connoramoreno:
Tests are now passing

parastell/utils.py
```diff
@@ -202,3 +207,119 @@ def smooth_matrix(matrix, steps, sigma):
...
+        # dim 0: vertex
+        # dim 1: edge
+        # dim 2: surface
+        # dim 3: volume
+        # dim 4: group
+        max_ids = {dim: 0 for dim in range(5)}
+        root_set = self.mb.get_root_set()
+        geom_sets = self.mb.get_entities_by_type_and_tag(
+            root_set, types.MBENTITYSET, self.geom_dim_tag, [None]
+        )
```
gonuke, line 253: Why not do this in a loop of `dim` and get all the sets of each dim at once and grab the max:
```
for dim in range(5):
    dim_sets = self.mb.get_entities_by_type_and_tag(root_set, types.MBENTITYSET, self.geom_dim_tag, [dim])
    set_ids = self.mb.tag_get_data(self.global_id_tag, dim_sets)
    max_ids[dim] = max(set_ids)
```

parastell/utils.py
```diff
@@ -202,3 +207,119 @@ def smooth_matrix(matrix, steps, sigma):
...
+        """Renumber entities, starting from the offset value for that
+            dimension.
+
+        Arguments:
+            offsets (dict): {entity dimension (int): id_offset (int)}
+        """
+        root_set = self.mb.get_root_set()
+        geom_sets = self.mb.get_entities_by_type_and_tag(
+            root_set, types.MBENTITYSET, self.geom_dim_tag, [None]
+        )
```
gonuke, line 272: This method may also be simplified if you can act on each dimension at a time.

parastell/utils.py
```diff
@@ -202,3 +207,119 @@ def smooth_matrix(matrix, steps, sigma):
...
+        ) as temp_file:
+            temp_filename = temp_file.name
+        merger.write_file(temp_filename)
+        temp_files.append(temp_filename)
+
+    merged_mbc = core.Core()
+    for file in temp_files:
+        merged_mbc.load_file(file)
+
+    Path.unlink(temp_filename)
```
gonuke, line 324: Don't you have to unlink all the files in `temp_files`?

parastell/utils.py
```diff
@@ -202,3 +207,119 @@ def smooth_matrix(matrix, steps, sigma):
...
+        merger.reassign_global_ids(max_ids)
+        max_ids = merger.get_max_ids()
+        with tempfile.NamedTemporaryFile(
+            delete=False, suffix=".h5m"
+        ) as temp_file:
+            temp_filename = temp_file.name
+        merger.write_file(temp_filename)
+        temp_files.append(temp_filename)
+
+    merged_mbc = core.Core()
```
gonuke, line 320: I think you can instantiate this MOAB core at the beginning of the method and just reuse the same temp_file, loading it into this MOAB core each time.

parastell/utils.py
```diff
@@ -202,3 +207,119 @@ def smooth_matrix(matrix, steps, sigma):
...
+
+        for entity_set in geom_sets:
+            dim = self.mb.tag_get_data(self.geom_dim_tag, entity_set)[0][0]
+            gid = self.mb.tag_get_data(self.global_id_tag, entity_set)[0][0]
+            max_ids[dim] = max(max_ids[dim], gid)
+
+        return max_ids
+
+    def reassign_global_ids(self, offsets):
+        """Renumber entities, starting from the offset value for that
```
gonuke, line 263: If you're going to renumber them anyway, do you want to ensure they are contiguous at the same time?  This method just shifts them all by some amount, but doesn't necessarily collapse them to a contiguous set.  This may not matter, but it can't hurt.

gonuke: Also, you could update `max_ids` as you go and return it as an output from this method and not have to loop through all the elements again each time you add more.

parastell/utils.py
```diff
@@ -202,3 +207,119 @@ def smooth_matrix(matrix, steps, sigma):
         previous_matrix = smoothed_matrix
 
     return smoothed_matrix
+
+
+class DAGMCMerger:
```
gonuke, line 212: This might be better name `DAGMCRenumber` ??  It doesn't actually do any merging and just handles renumbering

parastell/invessel_build.py
```diff
@@ -256,6 +302,150 @@ def generate_components(self):
...
+                    corner2 = rib.mb_verts[rib_pt_index + 1]
+                    corner3 = next_rib.mb_verts[rib_pt_index + 1]
+                    corner4 = next_rib.mb_verts[rib_pt_index]
+                    corners = [corner1, corner2, corner3, corner4]
+                    mb_tris += create_moab_tris_from_verts(
+                        corners,
+                        self.dag_model.mb,
+                        reverse=True if i == 0 else False,
+                    )
+            dagmc_surface = dagmc.Surface.create(self.dag_model)
```
gonuke, line 355: `DAGModel` provides an interface for this
```suggestion
            dagmc_surface = self.dag_model.create_surface()
```

parastell/invessel_build.py
```diff
@@ -256,6 +302,150 @@ def generate_components(self):
...
+        """Generate MBVERTEX entities from rib loci in all surfaces"""
+        [
+            surface.generate_pymoab_verts(self.mbc)
+            for surface in self.Surfaces.values()
+        ]
+
+    def generate_curved_surfaces_pydagmc(self):
+        """Generate the faceted representation of each curved surface and
+        add it to the PyDAGMC model, remembering the surface ids."""
+        curved_surface_ids = []
```
gonuke, line 340: Why not just used `self.curved_surface_ids` instead of this local list variable

parastell/invessel_build.py
```diff
@@ -256,6 +302,150 @@ def generate_components(self):
...
+                    )
+            dagmc_surface = dagmc.Surface.create(self.dag_model)
+            self.dag_model.mb.add_entities(dagmc_surface.handle, mb_tris)
+            curved_surface_ids.append(dagmc_surface.id)
+        self.curved_surface_ids = curved_surface_ids
+
+    def generate_end_cap_surfaces_pydagmc(self):
+        """Generate the faceted representation of the planar end cap surfaces
+        and add them to the PyDAGMC model, remembering the surface ids."""
+        end_cap_surface_ids = []
```
gonuke, line 363: why not just use the member variable `self.end_cap_surface_ids` instead of this local list variable?

parastell/invessel_build.py
```diff
@@ -256,6 +302,150 @@ def generate_components(self):
...
+        for surface, next_surface in zip(
+            list(self.Surfaces.values())[0:-1],
+            list(self.Surfaces.values())[1:],
+        ):
+            mb_tris = self.connect_ribs_with_tris_moab(
+                surface.Ribs[0], next_surface.Ribs[0]
+            )
+            end_cap_start = dagmc.Surface.create(self.dag_model)
```
gonuke, line 371:
```suggestion
            end_cap_start = self.dag_model.create_surface()
```

parastell/invessel_build.py
```diff
@@ -256,6 +302,150 @@ def generate_components(self):
...
+            self.mbc.add_entities(end_cap_start.handle, mb_tris)
+
+            mb_tris = self.connect_ribs_with_tris_moab(
+                surface.Ribs[-1], next_surface.Ribs[-1], reverse=True
+            )
+            end_cap_end = dagmc.Surface.create(self.dag_model)
```
gonuke, line 377:
```suggestion
            end_cap_end = self.dag_model.create_surface()
```

parastell/invessel_build.py
```diff
@@ -256,6 +302,150 @@ def generate_components(self):
...
+
+    def generate_end_cap_surfaces_pydagmc(self):
+        """Generate the faceted representation of the planar end cap surfaces
+        and add them to the PyDAGMC model, remembering the surface ids."""
+        end_cap_surface_ids = []
+        for surface, next_surface in zip(
+            list(self.Surfaces.values())[0:-1],
+            list(self.Surfaces.values())[1:],
+        ):
+            mb_tris = self.connect_ribs_with_tris_moab(
```
gonuke, line 368: put this in a loop over (0,-1) to do the two surfaces?

parastell/invessel_build.py
```diff
@@ -256,6 +302,150 @@ def generate_components(self):
...
+        ]
+
+    def generate_curved_surfaces_pydagmc(self):
+        """Generate the faceted representation of each curved surface and
+        add it to the PyDAGMC model, remembering the surface ids."""
+        curved_surface_ids = []
+        for i, surface in enumerate(self.Surfaces.values()):
+            mb_tris = []
+            for rib, next_rib in zip(surface.Ribs[0:-1], surface.Ribs[1:]):
+                for rib_pt_index, _ in enumerate(rib.rib_loci[0:-1]):
```
gonuke, line 344: for readability, I recommend that this loop variable be named the same as in `connect_ribs_with_tris_moab`

gonuke: actually - you should just call `connect_ribs_with_tris_moab` here...

gonuke (changes requested):
This is great use of PyDAGMC and pyMOAB!! It all looks pretty good.  I see a few places to streamline and simplify things still.

parastell/invessel_build.py
```diff
@@ -18,6 +20,38 @@
...
+            types.MBTRI, [corners[2], corners[1], corners[0]]
+        )
+        tri_2 = mbc.create_element(
+            types.MBTRI, [corners[3], corners[2], corners[0]]
+        )
+
+    return [tri_1, tri_2]
+
+
+export_allowed_kwargs = ["export_cad_to_dagmc", "dagmc_filename"]
```
connoramoreno, line 52: This can be removed since IVB CAD-to-DAGMC functionality was removed in a previous PR. I'm guessing this made its way back in during your rebase.

parastell/invessel_build.py
```diff
@@ -256,6 +302,150 @@ def generate_components(self):
...
+                list(self.dag_model.surfaces_by_id.keys())[-2:]
+            )
+        self.end_cap_surface_ids = end_cap_surface_ids
+
+    def generate_volumes_pydagmc(self):
+        """Use the curved surface and end cap surface IDs to build the
+        the volumes by applying the correct surface sense to each surface."""
+        # make the volumes and apply surface sense
+        for i in range(len(self.Surfaces) - 1):
+            dagmc.Volume.create(self.dag_model)
```
connoramoreno, line 389: There are typically more Pythonic alternatives to the `range(len())` pattern
```suggestion
        for i, _ in enumerate(self.Surfaces[:-1]):
            dagmc.Volume.create(self.dag_model)
```

parastell/invessel_build.py
```diff
@@ -256,6 +302,150 @@ def generate_components(self):
...
+                self.dag_model.surfaces_by_id[surface_id].surf_sense = [
+                    self.dag_model.volumes_by_id[i],
+                    self.dag_model.volumes_by_id[i + 1],
+                ]
+            # if it the last surface it goes to the implicit complement
+            else:
+                self.dag_model.surfaces_by_id[surface_id].surf_sense = [
+                    self.dag_model.volumes_by_id[i],
+                    None,
+                ]
```
connoramoreno, line 408: Not critical, but I think it may make more sense logically/be more readable to switch these `elif` and `else` statements, instead using
```suggestion
            # if it the last surface it goes to the implicit complement
            elif i != len(self.curved_surface_ids) - 1:
                self.dag_model.surfaces_by_id[surface_id].surf_sense = [
                    self.dag_model.volumes_by_id[i],
                    None,
                ]
            else:
                self.dag_model.surfaces_by_id[surface_id].surf_sense = [
                    self.dag_model.volumes_by_id[i],
                    self.dag_model.volumes_by_id[i + 1],
                ]
```

parastell/invessel_build.py
```diff
@@ -256,6 +302,150 @@ def generate_components(self):
...
+        if np.isclose(
+            (self._repeat + 1) * self.radial_build.toroidal_angles[-1], 360
+        ):
+            e = AssertionError(
+                "PyDAGMC workflow does not support modeling of the full 360 "
+                "degrees. Please consider modeling only one period. i.e. set "
+                "'repeat = 0'"
+            )
+            self._logger.error(e.args[0])
+            raise e
```
connoramoreno, line 441: We should probably be considerate of single-period stellarator folks. Although, would it be difficult to support 360-degree modeling (perhaps in a later PR)?
```suggestion
        if np.isclose(
            (self._repeat + 1) * self.radial_build.toroidal_angles[-1], 360
        ):
            e = AssertionError(
                "The PyDAGMC workflow does not support modeling 360-degree"
                "geometries. For configurations with more than one period, please "
                "consider modeling only one period. i.e. set 'repeat = 0'."
            )
            self._logger.error(e.args[0])
            raise e
```

parastell/invessel_build.py
```diff
@@ -256,6 +302,150 @@ def generate_components(self):
...
+        self._logger.info("Generating ivb dagmc model with pydagmc...")
```
connoramoreno, line 442:
```suggestion
        self._logger.info("Generating DAGMC model of in-vessel components with PyDAGMC...")
```

parastell/parastell.py
```diff
@@ -488,6 +497,15 @@ def export_cad_to_dagmc(
...
         )
+        if self.use_pydagmc:
+            self.magnet_model = export_path
+
+    def merge_magnet_and_ivb_dagmc_models(self):
+        magnet_mbc = core.Core()
+        magnet_mbc.load_file(str(self.magnet_model))
+        self.merged_model = merge_dagmc_files(
+            [self.invessel_build.dag_model.mb, magnet_mbc]
+        )
```
connoramoreno, line 508: I'm not sure there's much of a use-case for this to be a user-called function, since we know if a merged DAGMC model needs to be created based on whether an IVB and magnet set are created, and if PyDAGMC is being used. I think we should automate this.

parastell/utils.py
connoramoreno, line 212: Less of a suggestion and more of a question
```suggestion
class DAGMCMerger(object):
```

parastell/utils.py
```diff
@@ -202,3 +207,119 @@ def smooth_matrix(matrix, steps, sigma):
...
+        """Save to file.
+
+        Arguments:
+            filename (str): Path to save the file to. All filetypes supported
+                by MOAB are available. Write to h5m for use with DAGMC.
+        """
+        self.mb.write_file(filename)
+
+
+def merge_dagmc_files(models_to_merge):
```
connoramoreno, line 295: Doesn't seem like the I/O stream here actually uses any files
```suggestion
def merge_dagmc_models(models_to_merge):
```

parastell/utils.py
```diff
@@ -202,3 +207,119 @@ def smooth_matrix(matrix, steps, sigma):
...
+    def __init__(self, mb=None):
+        self.mb = mb
+        if mb is None:
+            self.mb = core.Core()
+        self.geom_dim_tag = self.mb.tag_get_handle(
+            "GEOM_DIMENSION", 1, types.MB_TYPE_INTEGER, types.MB_TAG_DENSE
+        )
+        self.global_id_tag = self.mb.tag_get_handle(
+            "GLOBAL_ID", 1, types.MB_TYPE_INTEGER, types.MB_TAG_DENSE
+        )
```
connoramoreno, line 228: Good practice to use a setter here

connoramoreno (changes requested):
Nice work here, excited to have this feature added. I think the added code design is pretty solid in general. I mostly have some additional code design and formatting suggestions.

parastell/invessel_build.py
connoramoreno: Ope, forgot to update the condition - good catch!

parastell/utils.py
Edgar-21: as I was implementing this I realized the whole thing can be much simpler as it doesn't hurt anything to have duplicate ids in the same pymoab core instance so long as they wind up being unique prior to saving the file. The new method is to just load all the models, then renumber entities from 1 to N where N is the number of entities, which ensures the ids are unique and contiguous.

parastell/utils.py
Edgar-21: renamed to DAGMCRenumberizer

parastell/invessel_build.py
Edgar-21: 🤯

parastell/invessel_build.py
Edgar-21: I don't think i would be too hard to do, just would need to make sure we handle the matching vertices at the start and end, as well as conditionally not making the end caps. I'd prefer to leave that to a future PR.

parastell/invessel_build.py
Edgar-21: Going with @gonuke's suggestion since I also like keeping them in order from inside out.

parastell/utils.py
Edgar-21: they are the same, classes inherit from object by default. I'll make the change though to avoid future confusion for anyone

Edgar-21:
Thanks for the feedback @gonuke and @connoramoreno, its ready for round two.

parastell/utils.py
```diff
@@ -202,3 +208,100 @@ def smooth_matrix(matrix, steps, sigma):
...
+            filename (str): Path to DAGMC model to be loaded.
+        """
+        self.mb.load_file(filename)
+
+    def renumber_ids(self):
+        """Renumbers the ids from 1 to N where N is the total number of
+        entities in that category.
+        """
+        for category in self.categories:
+            root_set = self.mb.get_root_set()
```
gonuke, line 270: You only need to do this once outside of the loop

parastell/utils.py
```diff
@@ -202,3 +208,100 @@ def smooth_matrix(matrix, steps, sigma):
...
+        mb (PyMOAB Core): PyMOAB core with a DAGMC models to be renumbered
+        loaded. Optional.
+    """
+
+    def __init__(self, mb=None):
+        self.mb = mb
+        if mb is None:
+            self.mb = core.Core()
+
+        self.categories = ["Vertex", "Curve", "Surface", "Volume", "Group"]
```
gonuke, line 226: This could be a class variable rather than a member variable

parastell/utils.py
```diff
@@ -202,3 +208,100 @@ def smooth_matrix(matrix, steps, sigma):
...
+        """
+        self.mb.load_file(filename)
+
+    def renumber_ids(self):
+        """Renumbers the ids from 1 to N where N is the total number of
+        entities in that category.
+        """
+        for category in self.categories:
+            root_set = self.mb.get_root_set()
+            category_set = self.mb.get_entities_by_type_and_tag(
```
gonuke, line 271: I'm OK with you using the category tag, but can't you also use the dimension tag for this?

parastell/utils.py
```diff
@@ -202,3 +208,100 @@ def smooth_matrix(matrix, steps, sigma):
...
+        return self.mb.tag_get_handle(
+            types.CATEGORY_TAG_NAME,
+            types.CATEGORY_TAG_SIZE,
+            types.MB_TYPE_OPAQUE,
+            types.MB_TAG_SPARSE,
+            create_if_missing=True,
+        )
+
+    @cached_property
+    def geom_dim_tag(self):
```
gonuke, line 248: seems like you don't use the dim tag any more... see note below...

parastell/utils.py
```diff
@@ -202,3 +208,100 @@ def smooth_matrix(matrix, steps, sigma):
...
+    def renumber_ids(self):
+        """Renumbers the ids from 1 to N where N is the total number of
+        entities in that category.
+        """
+        for category in self.categories:
+            root_set = self.mb.get_root_set()
+            category_set = self.mb.get_entities_by_type_and_tag(
+                root_set, types.MBENTITYSET, self.category_tag, [category]
+            )
+            set_ids = self.mb.tag_get_data(self.global_id_tag, category_set)
```
gonuke, line 274: Do you need to get `set_ids`?  Can't you use the length of `category_set`?

parastell/invessel_build.py
```diff
@@ -256,6 +308,144 @@ def generate_components(self):
...
+                    next_surface.Ribs[index],
+                    reverse=True if index == -1 else False,
+                )
+                end_cap = self.dag_model.create_surface()
+                self.mbc.add_entities(end_cap.handle, mb_tris)
+
+            self.end_cap_surface_ids.append(
+                list(self.dag_model.surfaces_by_id.keys())[-2:]
+            )
+
```
gonuke, line 377: Isn't it cleaner to append these one at a time as you build the surfaces?

parastell/invessel_build.py
```diff
@@ -256,6 +308,144 @@ def generate_components(self):
...
+    def _generate_curved_surfaces_pydagmc(self):
+        """Generate the faceted representation of each curved surface and
+        add it to the PyDAGMC model, remembering the surface ids."""
+        self.curved_surface_ids = []
+        for surf_idx, surface in enumerate(self.Surfaces.values()):
+            mb_tris = []
+            for rib, next_rib in zip(surface.Ribs[0:-1], surface.Ribs[1:]):
+                mb_tris += self._connect_ribs_with_tris_moab(
+                    rib, next_rib, reverse=True if surf_idx == 0 else False
```
gonuke, line 351:
```suggestion
                    rib, next_rib, reverse=(surf_idx == 0)
```

parastell/invessel_build.py
```diff
@@ -256,6 +308,144 @@ def generate_components(self):
...
+        self.end_cap_surface_ids = []
+        for surface, next_surface in zip(
+            list(self.Surfaces.values())[0:-1],
+            list(self.Surfaces.values())[1:],
+        ):
+            for index in (0, -1):
+                mb_tris = self._connect_ribs_with_tris_moab(
+                    surface.Ribs[index],
+                    next_surface.Ribs[index],
+                    reverse=True if index == -1 else False,
```
gonuke, line 369:
```suggestion
                    reverse=(index == -1)
```

parastell/invessel_build.py
gonuke: Note: you could do this without `surf_idx` is you made this condition `surface == self.Surfaces.values()[0]`.

To make it more readable, you could define:
```
surfaces = self.Surfaces.values()
first_surface = surfaces[0]
```
and then loop over `surfaces` without enumerate, and use the condition `surface == first_surface`

parastell/invessel_build.py
```diff
@@ -256,6 +308,144 @@ def generate_components(self):
...
+        # make the volumes and apply surface sense
+
+        [self.dag_model.create_volume() for _ in list(self.Surfaces)[:-1]]
+
+        # First surface goes to the implicit complement (plasma chamber)
+        self.dag_model.surfaces_by_id[
+            self.curved_surface_ids[0]
+        ].surf_sense = [
+            self.dag_model.volumes_by_id[1],
+            None,
```
gonuke, line 390: This order is correlated to having the triangles built with reverse sense on the first surface, right?  That is probably good to document in both places.

gonuke: Alternatively, you could build those triangles with a forward sense, and then set the sense here in the opposite order? or not?

parastell/invessel_build.py
```diff
@@ -256,6 +308,144 @@ def generate_components(self):
...
+        """Use the curved surface and end cap surface IDs to build the
+        the volumes by applying the correct surface sense to each surface."""
+        # make the volumes and apply surface sense
+
+        [self.dag_model.create_volume() for _ in list(self.Surfaces)[:-1]]
+
+        # First surface goes to the implicit complement (plasma chamber)
+        self.dag_model.surfaces_by_id[
+            self.curved_surface_ids[0]
+        ].surf_sense = [
```
gonuke, line 388: For readability it might be better if this was more like the blocks that follow:
```suggestion
        surface_id = self.curved_surface_ids[0]
        self.dag_model.surfaces_by_id[surface_id].surf_sense = [
```

parastell/invessel_build.py
```diff
@@ -256,6 +308,144 @@ def generate_components(self):
...
+                self.dag_model.volumes_by_id[surface_id],
+            ]
+
+        # if it the last surface it goes to the implicit complement
+        self.dag_model.surfaces_by_id[
+            self.curved_surface_ids[-1]
+        ].surf_sense = [
+            self.dag_model.volumes_by_id[self.curved_surface_ids[-1] - 1],
+            None,
+        ]
```
gonuke, line 405: As above:
```suggestion
        surface_id = self.curved_surface_ids[-1]
        self.dag_model.surfaces_by_id[surface_id].surf_sense = [
            self.dag_model.volumes_by_id[surface_id - 1],
            None
        ]
```

parastell/invessel_build.py
```diff
@@ -256,6 +308,144 @@ def generate_components(self):
...
+
+        # all end caps go to the implicit complement.
+        for i, end_cap_ids in enumerate(self.end_cap_surface_ids):
+            for id in end_cap_ids:
+                self.dag_model.surfaces_by_id[id].surf_sense = [
+                    self.dag_model.volumes_by_id[i + 1],
+                    None,
+                ]
```
gonuke, line 413:
```suggestion
        for volume_id, end_cap_ids in enumerate(self.end_cap_surface_ids, start=1):
            for surface_id in end_cap_ids:
                self.dag_model.surfaces_by_id[surface_id].surf_sense = [
                    self.dag_model.volumes_by_id[volume_id],
                    None,
                ]
```

parastell/invessel_build.py
```diff
@@ -256,6 +308,144 @@ def generate_components(self):
...
+            )
+
+    def _generate_volumes_pydagmc(self):
+        """Use the curved surface and end cap surface IDs to build the
+        the volumes by applying the correct surface sense to each surface."""
+        # make the volumes and apply surface sense
+
+        [self.dag_model.create_volume() for _ in list(self.Surfaces)[:-1]]
+
+        # First surface goes to the implicit complement (plasma chamber)
```
gonuke, line 385: What if we made end caps for the plasma and made it an actual volume?  It could make transport slightly more efficient?

gonuke (changes requested):
A few more comments, mostly in the process of building volumes

parastell/utils.py
Edgar-21: I was having a hard time getting the dimension tag to behave, I kept getting entity sets that were larger than I was expecting. Likely user error. I was looking at PyDAGMC to try to understand what I was doing wrong and noticed the category tag and it worked how I expected so I went with that.

parastell/invessel_build.py
Edgar-21: The plasma end caps are harder than the others, I tried connecting all the points to the magnetic axis, then when that didn't work (there wasn't line of sight from each point to the magnetic axis) I kind of just moved on. I am sure there are ways to close up the plasma chamber (some sort of tessellation algorithm or something like that) that I could try to implement, but figured that I would submit this first and tackle the plasma chamber stuff later.

parastell/utils.py
gonuke: No worries - we can definitely stick with that

parastell/invessel_build.py
Edgar-21: This is a list of lists, where each element has the end caps that correspond to a specific volume. I'll make a change to make it clearer what is happening.

parastell/invessel_build.py
Edgar-21: that is is a neat trick

parastell/invessel_build.py
Edgar-21: Yeah, I set the sense to be pointing at the implicit complement, or if the surface is between two volumes to be pointing out in the radial direction. The alternative also appears to work, though I would prefer to stick with keeping the sense pointing at the implicit complement if its all the same to you. I'll add some notes describing the convention.

connoramoreno (changes requested):
Relatively small review on my end. This is looking really good!

Edgar-21:
Rebased and ready for round 3.

connoramoreno (approved):
Looks great. Thanks for this addition and your hard work!

Edgar-21:
For reference using @connoramoreno's benchmark model we get results on the pydagmc geometry that are in agreement.


    TBR
        Cubit geometry: 1.1636 +/- 0.0012792
        CAD-to-DAGMC geometry: 1.16405 +/- 0.00129761
        PyDAGMC geometry: 1.16379 +/- 0.00161067
    Leakage fraction
        Cubit geometry: 0.48761 +/- 0.00100
        CAD-to-DAGMC geometry: 0.48819 +/- 0.00098
        PyDAGMC geometry: 0.48623 +/- 0.00086

gonuke (approved):
LGTM - great work to make this all happen

PR #191, opened 2025-02-04
add pydagmc to environment

Edgar-21:
Adds PyDAGMC to the testing environment. #190 depends on PyDAGMC.


connoramoreno (approved):
Looks good to me, thanks @Edgar-21

PR #192, opened 2025-02-04
Fix open-source DAGMC workflow

connoramoreno:
This branch modifies the open-source DAGMC workflow such that an in-memory `cadquery.Compound` object, comprising all IVB and magnet solids, is directly supplied directly to Gmsh, leveraging CAD-to-DAGMC's utility functions. This workflow skips imprinting, since it is not necessary for valid ParaStell geometries. This workflow also circumvents a merging routine since in-memory CAD solids are already "merged" (i.e., coincident surfaces are already shared between adjacent solids) due to the manner by which we create our IVB CAD solids (cutting interior components away).

### DIFF
--- a/parastell/parastell.py
+++ b/parastell/parastell.py
@@ -1,10 +1,9 @@
 import argparse
-import yaml
 from pathlib import Path
 
 import cubit
-import numpy as np
 import pystell.read_vmec as read_vmec
+import cadquery as cq
 import cad_to_dagmc
 
 from . import log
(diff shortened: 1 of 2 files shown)


Edgar-21:
Curious why some variables have the underscore `_geometry` and `_material_tags`? This indicates its a private variable or something?

connoramoreno:
> Curious why some variables have the underscore `_geometry` and `_material_tags`? This indicates its a private variable or something?

Yeah, that’s the idea. We follow that convention for most of our variables

gonuke (approved):
This looks OK to me, but I'm not a big fan of `cad_to_dagmc` API.  It doesn't seem like the function names are clear about what it actually does...

Are the tests of the cad-to-dagmc methods robust enough to rely on them to know this is working??

connoramoreno:
@gonuke CAD-to-DAGMC's test suite does not test these functions directly, but there is indirect testing since its `CadToDagmc.export_dagmc_h5m_file` method relies on the functions, and that is tested directly. We could add some tests on our side for peace of mind.

I could also PR into CAD-to-DAGMC to change the function names or wrap the functions with our own to give more transparent names.

Edgar-21:
We could add a test that runs overlap check or something to check the number of surfaces since that is what tipped us off to this behavior.

gonuke:
If you are all comfortable that it's working, I'll let one of you merge.  

We should discuss more robust testing at some point.

connoramoreno:
Let me benchmark running one of these models in a simulation against one made via Cubit before merging. I'll also document some geometry interrogation so we can be confident this functions as expected

connoramoreno:
Similarly to what @Edgar-21 has done in PR #190, I've added checks to `test_parastell` to ensure the proper number of surfaces and volumes are present in the generated DAGMC models

connoramoreno:
To benchmark the new CAD-to-DAGMC workflow against Cubit, I created a geometry with only the plasma chamber and a uniform 50-cm breeding volume using each workflow. The results as reported by OpenMC are as follows (30 batches of 10,000 particles), agreeing within uncertainty:

- TBR
    - Cubit geometry: 1.1636 +/- 0.0012792
    - CAD-to-DAGMC geometry: 1.16405 +/- 0.00129761
- Leakage fraction
    - Cubit geometry: 0.48761 +/- 0.00100
    - CAD-to-DAGMC geometry: 0.48819 +/- 0.00098

connoramoreno:
The checks that is not passing is doing so because of the format of this branch's name. Docker is getting confused about the "/". I was trying out a new convention since I learned it was "best practice" but I've learned my lesson not to follow it in the future

PR #195, opened 2025-02-10
Make Cubit installation optional

connoramoreno:
This branch isolates the use of Cubit within ParaStell such that it becomes an optional dependency. Cubit import occurs in a single location, the `init_cubit` function of the `cubit_utils.py` script (formerly named `cubit_io.py`), which is called only in functions using Cubit. As the radial distance utility is completely dependent on Cubit, no major modifications were made to that script. In addition, tests are modified to test Cubit functionality only if installed, including `test_parastell.py`.


parastell/magnet_coils.py
```diff
@@ -5,10 +5,15 @@
...
 
 from . import log
-from . import cubit_io as cubit_io
+from .cubit_utils import (
+    import_geom_to_cubit,
+    export_mesh_cubit,
+    mesh_volume_skeleton,
+    get_last_id,
+)
+from . import cubit_utils
```
connoramoreno, line 16: I'd like to note (for reviewers) that `cubit_utils` is imported separately here for its `initialized` variable. If `initialized` is imported into the namespace of `magnet_coils`, changes to the variable do not persist when modified by calls to the imported functions

connoramoreno:
I would also like to note that all tests indeed pass (other than those skipped) when Cubit is removed from `PYTHONPATH`

parastell/cubit_utils.py
```diff
@@ -196,3 +189,120 @@ def import_geom_to_cubit(filename, import_dir=""):
...
+        f"{min_size} max_size {max_size} max_gradient {max_gradient} "
+        "min_num_layers_3d 1 min_num_layers_2d 1 min_num_layers_1d 1"
+    )
+    cubit.cmd(f"mesh volume {volume_ids_str}")
+
+
+def get_last_id(entity):
+    """Returns the ID of the most recently created entity of a given type.
+
+    Arguments:
```
Edgar-21, line 302: might be nice to list valid strings here for anyone who is not familiar with the entity types in cubit.

Edgar-21:
Do we want to update the example parastell_cad_to_dagmc_example.py to not use cubit at all? Right now it calls cubit to make a volumetric mesh of the magnets.

parastell/cubit_utils.py
```diff
@@ -63,8 +64,6 @@ def export_step_cubit(filename, export_dir=""):
         export_dir (str): directory to which to export the STEP output file
             (defaults to empty string).
     """
-    init_cubit()
```
gonuke, line 66: I think we included these to be safe.  By removing them, I guess we require folks to be sure they have called `init_cubit()` or they'll get an error?

parastell/magnet_coils.py
gonuke: Should that be added as a comment?

gonuke (changes requested):
I haven't reviewed all the new tests yet, but here are a couple of little comments.

parastell/cubit_utils.py
connoramoreno: Yes, which I think makes sense for the functions that require certain actions within Cubit to have been taken (e.g., exports, imprinting, merging, etc.).

gonuke:
Waiting for a rebase following #196

README.md
```diff
@@ -5,24 +5,23 @@
...
 - [MOAB](https://bitbucket.org/fathomteam/moab/src/master/)
-- [Coreform Cubit](https://coreform.com/products/downloads/)
 - [CAD-to-DAGMC](https://github.com/fusion-energy/cad_to_dagmc)
 - [OpenMC](https://github.com/openmc-dev/openmc)
 - [NumPy](https://numpy.org/install/)
 - [SciPy](https://scipy.org/install/)
-- [scikit-learn](https://scikit-learn.org/stable/install.html)
 - [PyYAML](https://pyyaml.org/wiki/PyYAMLDocumentation)
-- [PyDAGMC](https://github.com/svalinn/pydagmc)
+- [Coreform Cubit](https://coreform.com/products/downloads/) (optional)
```
gonuke, line 24: Should we list some of these dependencies as optional?  Also, Is CAD-to-DAGMC optional if one has Cubit?

tests/test_parastell.py
```diff
@@ -29,28 +29,171 @@ def remove_files():
             Path.unlink(file)
 
 
+def check_cubit():
+    try:
+        import cubit
+
+        return True
+    except ImportError:
+        return False
```
gonuke, line 38: I wonder if it makes sense to put this in `cubit_utils`?  Not sure, to be honest, but seems like it could be useful...

tests/test_parastell.py
```diff
@@ -29,28 +29,171 @@ def remove_files():
...
+
+
 def check_surfaces_and_volumes(filename, num_surfaces_exp, num_volumes_exp):
     dagmc_model = dagmc.DAGModel(str(Path(filename).with_suffix(".h5m")))
 
     assert len(dagmc_model.surfaces) == num_surfaces_exp
     assert len(dagmc_model.volumes) == num_volumes_exp
 
 
+def create_ivb_cad_magnets_from_filaments(stellarator_obj):
```
gonuke, line 48: Maybe we should have docstrings or other comments to clarify what each test should actually be testing and what result is being used to accomplish the test

tests/test_parastell.py
```diff
@@ -29,28 +29,171 @@ def remove_files():
...
+    component_name_exp = "component"
+    radial_build_dict = {
+        component_name_exp: {
+            "thickness_matrix": np.ones(
+                (len(toroidal_angles), len(poloidal_angles))
+            )
+            * 10
+        }
+    }
+    num_ribs = 11
```
gonuke, line 101: This is reused across multiple tests. Can we define once? Maybe a test fixture?

gonuke (changes requested):
Looking for a little more documentation to help review tests

tests/test_parastell.py
connoramoreno: I think organizationally that makes sense

tests/test_parastell.py
connoramoreno: These `create_...` functions aren't tests, but since it's repeated across multiple functions, I agree it's probably best to define just once

tests/test_invessel_build.py
```diff
@@ -94,8 +96,10 @@ def test_ivb_basics(invessel_build):
     remove_files()
 
 
-def test_ivb_construction(invessel_build):
-
+def test_ivb_cadquery_construction(invessel_build):
+    """Tests whether the InVesselBuild CadQuery workflow functions as
+    expected.
```
gonuke, line 101: Might be good to include a little more detail, e.g.:
```suggestion
    expected, by testing whether:
       * the correct number of components were assembled
       * the first rib has the correct number of points
       * the rib is defined by floating point numbers
```

gonuke (changes requested):
I left a suggestion for a little more detail in one test, that could be included in each test.

gonuke (approved):
Thanks for this important update @connoramoreno

PR #196, opened 2025-02-10
Modify `utils.expand_list` to only conditionally add entries

connoramoreno:
Modifies `utils.expand_list` to add entries only when `num > len(list)`, and otherwise preserve the original list, which is the expected behavior. I discovered that if `num < len(list)` the resultant `list_exp` could be different than the input `list`.

### DIFF
--- a/parastell/utils.py
+++ b/parastell/utils.py
@@ -96,8 +96,8 @@ def expand_list(list, num):
     for entry, next_entry in zip(list[:-1], list[1:]):
         # Only add entries to current block if difference between entry and
         # next_entry is greater than desired average
-        num_new_entries = 0;
-        
+        num_new_entries = 0
+
         if next_entry - entry > avg_diff:
             num_new_entries = int(round(next_entry - entry / avg_diff))
 
--- a/parastell/utils.py
+++ b/parastell/utils.py
@@ -95,11 +95,21 @@ def expand_list(list, num):
     for entry, next_entry in zip(list[:-1], list[1:]):
         num_new_entries = int(round((next_entry - entry) / avg_diff))
 
-        # Don't append the last entry in the created linspace to avoid adding
-        # it twice when the next created linspace is appended
+        # Conditionally include the last entry in the created linspace to avoid
+        # adding it twice when the next created linspace is appended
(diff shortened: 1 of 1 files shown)


parastell/utils.py
```diff
@@ -95,11 +95,22 @@ def expand_list(list, num):
...
+        # In the event that num << len(list), num_new_entries = 0
+        if num_new_entries == 0:
+            # Returns an array with 'entry' as the only element
+            new_entries = np.linspace(
+                entry, next_entry, num=num_new_entries + 1
+            )
+        else:
+            new_entries = np.linspace(
+                entry, next_entry, num=num_new_entries + 1
+            )[:-1]
```
gonuke, line 109: if `num_new_entries == 1` doesnt' this result in no new entries?  this evaluates to only being the original `entry`?

gonuke (changes requested):
I think I've lost track a little with how this is supposed to work, but I wonder if you can do something logically simpler by manually appending `entry` and then appending  `np.linspace(entry, next_entry, num_new_entries+2)[1:-1]`?

When `num_new_entries < 1` this gives an empty list

parastell/utils.py
```diff
@@ -93,14 +94,25 @@ def expand_list(list, num):
...
-            list_exp,
-            np.linspace(entry, next_entry, num=num_new_entries + 1)[:-1],
-        )
+        # Only add entries to current block if difference between entry and
+        # next_entry is greater than desired average
+        num_new_entries = (next_entry - entry) / avg_diff
+        if num_new_entries >= 1.0:
+            num_new_entries = int(round(num_new_entries))
+        else:
+            num_new_entries = 0
```
gonuke, line 103: This seems slightly cleaner to me (?) and matches your comment a little better, perhaps
```suggestion
        num_new_entries = 0;
        
        if next_entry - entry > avg_diff:
            num_new_entries = int(round(next_entry - entry / avg_diff))
```

gonuke (changes requested):
One little tweak for readability...

gonuke (approved):
LGTM - thanks for bearing with the minor tweaks

PR #197, opened 2025-02-17
expand_list bug fix

Edgar-21:
I noticed that the new utils.expand_list() function was not behaving as I expected, and was throwing an error rather than expanding my list. I've added a test to check that the function functions here, demonstrating the bug and I will push a fix shortly.

### DIFF
--- a/parastell/utils.py
+++ b/parastell/utils.py
@@ -86,6 +86,9 @@ def expand_list(list_to_expand, num):
     Returns:
         list_exp (iterable of float): expanded list.
     """
+    if len(list_to_expand) >= num:
+        return list_to_expand
+
     list_exp = []
 
     init_entry = list_to_expand[0]
--- a/parastell/utils.py
+++ b/parastell/utils.py
@@ -72,14 +72,15 @@ def enforce_helical_symmetry(matrix):
     return matrix
 
 
-def expand_list(list, num):
+def expand_list(list_to_expand, num):
     """Expands a list of ordered floats to a total number of entries by
     linearly interpolating between entries, inserting a proportional number of
-    new entries between original entries. If num < len(list), no entries are
-    added.
+    new entries between original entries. If num <= len(list), list_to_expand
+    is not modified. It is possible that the result will have slightly more
+    or fewer elements than num, due to round off approximations.
 
     Arguments:
-        list (iterable of float): list to be expanded.
+        list_to_expand (iterable of float): list to be expanded.
         num (int): desired number of entries in expanded list.
 
     Returns:
@@ -87,24 +88,28 @@ def expand_list(list, num):
     """
     list_exp = []
 
-    init_entry = list[0]
-    final_entry = list[-1]
+    init_entry = list_to_expand[0]
+    final_entry = list_to_expand[-1]
     extent = final_entry - init_entry
 
     avg_diff = extent / (num - 1)
 
-    for entry, next_entry in zip(list[:-1], list[1:]):
+    for entry, next_entry in zip(list_to_expand[:-1], list_to_expand[1:]):
         # Only add entries to current block if difference between entry and
         # next_entry is greater than desired average
         num_new_entries = 0
 
         if next_entry - entry > avg_diff:
+            # Goal is to create bins of approximately avg_diff width between
+            # entry and next_entry
             num_new_entries = int(round((next_entry - entry) / avg_diff)) - 1
 
         # Manually append first entry
         list_exp = np.append(list_exp, entry)
 
         # If num_new_entries == 0, don't add new entries
+        # First and last elements of new_entries are entry and next_entry,
+        # respectively
         new_entries = np.linspace(
             entry,
             next_entry,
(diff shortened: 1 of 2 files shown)


tests/test_utils.py
```diff
@@ -35,3 +35,29 @@ def test_dagmc_renumbering():
     assert len(combined_model.volumes) == num_vol_exp
     assert max(combined_model.volumes_by_id.keys()) == max_vol_id_exp
     assert all(mat in mats_exp for mat in mats)
+
+
+def test_expand_list():
```
connoramoreno, line 40: This test looks great. Can you include a docstring for this test in a similar format to what we have for the other tests?

parastell/utils.py
```diff
@@ -99,7 +99,7 @@ def expand_list(list, num):
         num_new_entries = 0
 
         if next_entry - entry > avg_diff:
-            num_new_entries = int(round(next_entry - entry / avg_diff))
+            num_new_entries = int(round((next_entry - entry) / avg_diff)) - 1
```
connoramoreno, line 102: Can you include a comment about why this `- 1` is necessary?

connoramoreno (changes requested):
Thanks for this fix @Edgar-21! Just some small requests concerning documentation

parastell/utils.py
gonuke: Or do we need to use `floor` instead of `round`?

gonuke: I wonder if we construct a test where `(next_entry - entry) = avg_diff * 1.1` vs a test where `(next_entry - entry) = avg_diff * 1.8`?  (3.1 vs 3.8)

connoramoreno: If you mean `floor` instead of `round(...) - 1`, then I think the latter is the way to go. In the case that `avg_diff` evenly divides `next_entry - entry` (i.e., results in a whole number), `floor` will have the same issue that `int(round(...))` does and need to be subtracted by one.

If you mean `floor(...) - 1` instead of `round(...) - 1`, then I think comes down to a design choice. `floor` will bias towards fewer array elements, while `round` will always give the closest possible to the user input, depending on the integer math.

Edgar-21: The point of subtracting 1 is that we want to make 'bins' that are near to avg_diff in width, so if we find we can split the distance between `entry` and `next_entry` into 2 bins, we want to add one new element to the list, and so on. Unfortunately it is still possible to get an expanded list out that does not quite have the requested number of elements. I am ok with that behavior, as we now have consistent behavior when providing `num` that is less than or equal to the length of the list to be expanded.

I'll push a test that shows the case where we don't get the appropriate number of elements and we can decide how we want to proceed, whether that be modifying the test, or the code to get the test to pass.

connoramoreno: Sorry @Edgar-21, I meant include a comment in documentation for why we subtract one, unless you don't think it can be explained concisely enough for in-code documentation.

tests/test_utils.py
```diff
@@ -35,3 +35,37 @@ def test_dagmc_renumbering():
...
+    """Tests utils.expand_list() to ensure returned arrays are the length
+    expected, and contain the expected values."""
```
connoramoreno, line 42:
```suggestion
    """Tests utils.expand_list() to ensure returned arrays are the length
    expected, and contain the expected values, by testing if:
        * the expected entries are added to uniformly and non-uniformly spaced
          lists
        * entries are added when the requested size is less than or equal to
          that of the input list (no entries should be added)
    """
```

connoramoreno (changes requested):
Some comments clarifying my previous requests about documentation

parastell/utils.py
connoramoreno: I'm also okay with the current behavior, since we're operating within the limits of integer math. I'm unsure of a way to ensure the requested number of entries is always created while still including all entries from the original list. I say we modify the test to reflect current behavior.

Edgar-21: I did add a comment to the code as well, lines 103 to 104. Figured I'd leave a more detailed explanation as well where I'm not limited by trying to be concise. It's not my favorite comment I've ever left but I was having a hard time making a short explanation.

tests/test_utils.py
Edgar-21: I get what you mean now about documentation like the other tests, thanks for the suggestion

Edgar-21:
The same round off behavior that makes it difficult to get exactly the number of elements requested also makes relying on internal logic not to modify the list when `num <= len(list_to_expand)` inconsistent when dealing with non-uniform spacing of list elements. I added a check at the beginning that ensures that `list_to_expand` does not get modified unless `num>len(list_to_expand)`

parastell/utils.py
Edgar-21: updated tests to have this be expected behavior

connoramoreno (approved):
Thanks for this fix @Edgar-21, LGTM. I'll delay merging until @gonuke also approves.

gonuke (approved):
LGTM - thanks @Edgar-21

Issue #108, opened 2024-05-28
Support alternative geometric definitions of the first wall


gonuke:
Currently, the first wall is defined by evaluating a VMEC description of the plasma at some value of s > 1 (but not too much).

There are users interested in being able to provide an alternative definition of the first wall, including an (r,z,phi) cloud of points.

One known issue is to consider how users will want to logically define component thicknesses relative to such a different first wall definition.

PR #198, opened 2025-02-22
Support Custom FW Definitions

Edgar-21:
The approach I have taken is to create a class that has the same methods and behavior as those that we use from VMECData.

In this case, the custom first wall is defined in R-Z coordinates in a manner similar to how we define ribs. By assuming that the points on each rib are spaced evenly in the poloidal direction, a continuous surface which is a function of poloidal and toroidal angle can be created by interpolation of those points.

I was thinking we could set something up where there is a base class for representing the innermost surface, call it ReferenceSurface() or something along those lines, then we could have two classes that inherit from it, one that works with VMEC data and one that works with R-Z data, and by doing so we could have some infrastructure set up to possibly extend to other formats if it becomes necessary.

For right now I just added a vmec2xyz() method to my new KisslingerSurface() class that allows this class to be slotted in in place of a VMECData() object as a demonstration. I've included custom_geometry_example.py for a bit of context.

Addresses #108

### DIFF
--- a/parastell/invessel_build.py
+++ b/parastell/invessel_build.py
@@ -684,6 +684,8 @@ class Rib(object):
         if not np.all(self.offset_list == 0):
             self.rib_loci += self.offset_list[:, np.newaxis] * self._normals()
 
+        self.rib_loci[-1] = self.rib_loci[0]
+
     def _generate_pymoab_verts(self, mbc):
         """Converts point-loci to MBVERTEX and adds them to a PyMOAB
         Core instance. The first and last rib loci are identical. To avoid
--- a/parastell/utils.py
+++ b/parastell/utils.py
@@ -314,7 +314,22 @@ def combine_dagmc_models(models_to_merge):
     return dagmc.DAGModel(renumberizer.mb)
 
 
-def ribs_from_kisslinger_format(filename, start_line=3, scale=m2cm):
+def rotate_ribs(ribs, angle):
+    """Rotate a set of NxMx3 set of loci about the Z axis in the
+    counter-clockwise direction"""
+    angle = np.deg2rad(angle)
+    rotation_mat = np.array(
+        [
+            [np.cos(angle), -np.sin(angle), 0],
+            [np.sin(angle), np.cos(angle), 0],
+            [0, 0, 1],
+        ]
+    )
+    ribs = np.dot(ribs, rotation_mat.T)
+    return ribs
+
+
+def ribs_from_kisslinger_format(filename, start_line=2, scale=m2cm):
     """Reads a Kisslinger format file and returns a list of toroidal angles,
     along with a list of lists of the rib loci (in x,y,z) at each toroidal
     angle. It is expected that toroidal angles are provided in degrees.
@@ -322,8 +337,7 @@ def ribs_from_kisslinger_format(filename, start_line=3, scale=m2cm):
     Arguments:
         filename (str): Path to the file to be read.
         start_line (int): Line at which the data should start being read,
-            should skip any comments and the line describing the number of
-            toroidal angles, poloidal angles, and periods. Defaults to 3.
+            should skip any comments. Defaults to 2.
         scale (float): Amount to scale the r-z coordinates by. Defaults to 100.
     Returns:
         toroidal_angles (list of float): List of all the toroidal angles in the
@@ -342,7 +356,11 @@ def ribs_from_kisslinger_format(filename, start_line=3, scale=m2cm):
     toroidal_angles = []
 
     for line in data[start_line - 1 :]:
-        if "\t" not in line:
+        if line.count("\t") == 2:
+            num_toroidal_angles, num_poloidal_angles, periods = (
+                line.rstrip().split("\t")
+            )
+        elif "\t" not in line and "#" not in line:
             toroidal_angle = float(line.rstrip())
             toroidal_angles.append(toroidal_angle)
             if len(profile) != 0:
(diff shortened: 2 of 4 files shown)


parastell/utils.py
```diff
@@ -311,3 +312,178 @@ def combine_dagmc_models(models_to_merge):
...
+
+    with open(file=filename) as file:
+        data = file.readlines()
+
+    profiles = []
+    profile = []
+    toroidal_angles = []
+
+    for line in data[start_line - 1 :]:
+        if line.count("\t") == 2:
```
gonuke, line 359: searching for tabs makes me nervous.... does T1E have a standard format/reader for these files?  Is there a rigid structure we can rely upon?

parastell/utils.py
```diff
@@ -311,3 +312,178 @@ def combine_dagmc_models(models_to_merge):
...
+    for phi, rib in zip(toroidal_angles, ribs):
+        for theta, rib_locus in zip(poloidal_angles, rib):
+            x_data.append(rib_locus[0])
+            y_data.append(rib_locus[1])
+            z_data.append(rib_locus[2])
+            grid_points.append([phi, theta])
+    return x_data, y_data, z_data, grid_points
+
+
+class KisslingerSurface(object):
```
gonuke, line 399: Should this be in IVB somewhere as a variation on the current surface definition to allow us to use them more interchangeably?  I haven't really thought about the details, but assumed that's how this would come together, similar to what you've done with magnet definitions

gonuke (commented):
Some high level design questions...

I think my thought is similar to what you describe in the comments for the PR, but doesn't really show up in the code yet.

parastell/utils.py
connoramoreno: Perhaps it would be better to input rib loci (shape = (num_ribs, num_rib_pts, 3)), instead of reading them from a file? Relying on the user to extract data from the file removes any reliance of this method on a specific file format.

Edgar-21: There is some structure, I got the code snippet that writes the files from Bonita today and it specifically uses `'\t'` as a delimiter. I could make the delimiter character a variable perhaps. Or we could leave it up to users to have their own function that reads their data. The class I have created to make custom FW takes a set of points defining ribs as an input so as to make it function with any sort of first wall definition that can be represented with ribs.

gonuke: It may use tabs, that's fine, but I expect that the structure is defined by the order in which certain lines appear and the data expected to be on each line.

gonuke: If I reverse engineer what you wrote, I'm guessing the format is:
```
num_toroidal_angles num_poloidal_angles periods
# repeat for each toroidal angle
toroidal_angle
### repeat for each poloidal angle
r z
```
If so, then you can write a parser that
1. Reads the first line to get those three values
2. For loop over num toroidal angles
    1. read toroidal angle
    2. Initialize `profile`
    3. For loop over poloidal angles
        1. Read `r` & `z`
        2. Convert to `x,y,z`
        3. Append to `profile`
    4. Append to `profiles`

parastell/invessel_build.py
```diff
@@ -54,16 +60,256 @@ def create_moab_tris_from_verts(corners, mbc, reverse=False):
...
+        # Toroidal Periodicity
+        rotated_ribs = rotate_ribs(self.rib_data, -max(self.toroidal_angles))[
+            0:-1
+        ]
+        shifted_toroidal_angles = self.toroidal_angles[0:-1] - max(
+            self.toroidal_angles
+        )
+        self._extract_rib_data(
+            rotated_ribs, shifted_toroidal_angles, self.poloidal_angles
+        )
```
gonuke, line 206:
```suggestion
        toroidal_shift = -max(self.toroidal_angles)
        rotated_ribs = rotate_ribs(self.rib_data, toroidal_shift)
        shifted_toroidal_angles = self.toroidal_angles + toroidal_shift
        self._extract_rib_data(
            rotated_ribs[0:-1], shifted_toroidal_angles[0:-1], self.poloidal_angles
        )
```

parastell/invessel_build.py
```diff
@@ -54,16 +60,256 @@ def create_moab_tris_from_verts(corners, mbc, reverse=False):
...
+
+        # Poloidal Periodicity
+        shifted_poloidal_angles = self.poloidal_angles[0:-1] - max(
+            self.poloidal_angles
+        )
+        rib_subset = self.rib_data[:, 0:-1, :]
+        self._extract_rib_data(
+            rib_subset, self.toroidal_angles, shifted_poloidal_angles
+        )
```
gonuke, line 215:
```suggestion
        poloidal_shift = -max(self.poloidal_angles)
        shifted_poloidal_angles = self.poloidal_angles + polodial_shift
        self._extract_rib_data(
            self.rib_data[:, 0:-1, :], self.toroidal_angles, shifted_poloidal_angles[0:-1]
        )
```

parastell/invessel_build.py
```diff
@@ -54,16 +60,256 @@ def create_moab_tris_from_verts(corners, mbc, reverse=False):
...
+        ]
+        shifted_toroidal_angles = self.toroidal_angles[1:] + max(
+            self.toroidal_angles
+        )
+
+        self._extract_rib_data(
+            rotated_ribs,
+            shifted_toroidal_angles,
+            self.poloidal_angles,
+        )
```
gonuke, line 236:
```suggestion
        toroidal_shift = max(self.toroidal_angles)
        rotated_ribs = rotate_ribs(self.rib_data, toroidal_shift)
        shifted_toroidal_angles = self.toroidal_angles + toroidal_shift
        self._extract_rib_data(
            rotated_ribs[1:], shifted_toroidal_angles[1:], self.poloidal_angles
        )
```

parastell/invessel_build.py
```diff
@@ -54,16 +60,256 @@ def create_moab_tris_from_verts(corners, mbc, reverse=False):
...
+
+        # Poloidal Periodicity
+        shifted_poloidal_angles = self.poloidal_angles[1:] + max(
+            self.poloidal_angles
+        )
+        rib_subset = self.rib_data[:, 1:, :]
+        self._extract_rib_data(
+            rib_subset, self.toroidal_angles, shifted_poloidal_angles
+        )
```
gonuke, line 245:
```suggestion
        poloidal_shift = max(self.poloidal_angles)
        shifted_poloidal_angles = self.poloidal_angles + polodial_shift
        self._extract_rib_data(
            self.rib_data[:, 1:, :], self.toroidal_angles, shifted_poloidal_angles[1:]
        )
```

parastell/invessel_build.py
```diff
@@ -54,16 +60,256 @@ def create_moab_tris_from_verts(corners, mbc, reverse=False):
...
+        rib_data, toroidal_angles, and poloidal_angles. Adds copies of the data
+        shifted by one period ahead of and behind provided data in the toroidal
+        and poloidal directions to preserve periodicity.
+        """
+        self.x_data = []
+        self.y_data = []
+        self.z_data = []
+        self.grid_points = []
+
+        # Toroidal Periodicity
```
gonuke, line 197: Does it matter what order things are added to these for the interpolation functions to work?

gonuke (changes requested):
I like how this is coming together, but have a few thoughts for simplification and readability

parastell/invessel_build.py
```diff
@@ -54,16 +60,256 @@ def create_moab_tris_from_verts(corners, mbc, reverse=False):
...
+        toroidal_angles = np.rad2deg(toroidal_angles)
+        poloidal_angles = np.rad2deg(poloidal_angles)
+        for toroidal_angle, poloidal_angle in zip(
+            toroidal_angles, poloidal_angles
+        ):
+            x = self.x_interp(toroidal_angle, poloidal_angle)
+            y = self.y_interp(toroidal_angle, poloidal_angle)
+            z = self.z_interp(toroidal_angle, poloidal_angle)
+            coord = np.array([x, y, z])
+            # need to force the point to lie directly on the plane for
```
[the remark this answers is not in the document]
Edgar-21: I think we only know the ratio of x to y, not the actual values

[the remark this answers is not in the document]
gonuke: Of course, 🤦‍♂️. Could you tabulate/interpolate on `r` instead of `x` & `y`?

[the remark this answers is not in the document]
Edgar-21: The RibBasedSurface object is set up to work with a point cloud of Cartesian coordinates currently, but we could switch it to work with R-Z-toroidal angle instead, and I could see how that might end up putting the interpolated values closer to the correct plane, probably worth checking out.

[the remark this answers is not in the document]
Edgar-21: I see now that if we interpolate in R-Z we can then just use the toroidal angle provided directly instead of hoping the point winds up on the plane from the interpolator, I like that a lot better, I'll modify things accordingly

parastell/invessel_build.py
Edgar-21: I don't think so, it just made sense to me to add the data in order of increasing angle

parastell/invessel_build.py
Edgar-21: Made a change very similar to this but not exactly as when using R-Z interpolation there is no need to rotate ribs.

parastell/invessel_build.py
```diff
@@ -648,11 +894,12 @@ def _vmec2xyz(self, poloidal_offset=0):
                 poloidal angles for evaluating the location of the Cartesian
                 points (optional, defaults to 0).
         """
-        return self.scale * np.array(
-            [
-                self.vmec_obj.vmec2xyz(self.s, theta, self.phi)
-                for theta in (self.theta_list + poloidal_offset)
-            ]
+        toroidal_angles = np.ones(len(self.theta_list)) * self.phi
```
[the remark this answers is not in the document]
Edgar-21: I wanted to have the `ReferenceSurface.angles_to_xyz()` to accept two lists since I could see wanting to evaluate for multiple toroidal angles (maybe it would be useful for making discrete components or something, I'm not quite sure it just feels like a good idea).

Here are the options I see, though maybe there is something I have missed
a. Assume only one toroidal angle will be provided as you suggest
b. Implement some input checking and handling in `angles_to_xyz()` so single values, lists, or combinations thereof can be provided
c. leave as is

I'm happy to do whichever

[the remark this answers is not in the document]
gonuke: I prefer (a) for now until/unless we end something different

parastell/invessel_build.py
```diff
@@ -54,16 +60,256 @@ def create_moab_tris_from_verts(corners, mbc, reverse=False):
...
+            plane_norm = np.array(
+                [
+                    -np.sin(np.deg2rad(toroidal_angle)),
+                    np.cos(np.deg2rad(toroidal_angle)),
+                    0,
+                ]
+            )
+            delta_alpha = (plane_norm * coord).sum() / np.sqrt(
+                np.sum(np.square(plane_norm))
+            )
```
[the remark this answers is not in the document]
Edgar-21: using RZ interpolation makes this bit unecessary, it will be removed

gonuke (changes requested):
A couple of questions about the Kisslinger file format, including how we know the list of poloidal angles that appear to be necessary to create a `RibBasedSurface`?

Examples/custom_first_wall_profile_example.py
```diff
@@ -0,0 +1,102 @@
...
+custom_surface_rz_ribs = np.load(
+    "../tests/files_for_tests/custom_surface_rz_ribs.npy"
+)
+
+# For this example, the ribs and points on the ribs are evenly spaced, which
+# is not required.
+num_toroidal_angles, num_poloidal_angles, _ = custom_surface_rz_ribs.shape
+
+toroidal_angles = np.linspace(0, 90, num_toroidal_angles)
+poloidal_angles = np.linspace(0, 360, num_poloidal_angles)
```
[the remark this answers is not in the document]
Edgar-21: There isn't any data in the kisslinger file for poloidal angle, I've assumed them to be evenly spaced in the example Bonita shared with me and that seemed to work ok.

parastell/utils.py
```diff
@@ -389,3 +389,65 @@ def dagmc_volume_to_step(
...
+    profiles = []
+    toroidal_angles = []
+    num_toroidal_angles, num_poloidal_angles, periods = (
+        int(x) for x in data[0].rstrip().split(delimiter)
+    )
+
+    ribs = [
+        data[1:][x : x + num_poloidal_angles + 1]
+        for x in range(0, len(data[1:]), num_poloidal_angles + 1)
+    ]
```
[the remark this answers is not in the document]
Edgar-21: I think these are equivalent, I used `data[1:]` since we are interested in everything after the first line to determine the ribs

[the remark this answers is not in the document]
Edgar-21: seems to automatically mark as resolved when I commit a suggestion.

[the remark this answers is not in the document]
gonuke: Does this still work? - I assume so... but since I wasn't sure about the format, I wasn't sure how this would work.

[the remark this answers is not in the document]
Edgar-21: you have the format spot on above, and yes, it still does work with the suggested change

gonuke (commented):
I'm ready to approve this if it works, but wonder if we can make up a Kisslinger file to have a test of that?

PR #200, opened 2025-03-05
Add pyproject, and make pip installable. 

MicahGale:


pyproject.toml
```diff
@@ -0,0 +1,68 @@
...
+	"moab=5.5.1[build=nompi_tempest_*]",
+	"cad_to_dagmc >=0.8.2",
+	"ca-certificates",
+	"certifi",
+	"openssl",
+	"openmc>=0.15.0[build=dagmc_nompi_*]",
+	"matplotlib",
+	"netcdf4",
+	"pyyaml",
+	"git+https://github.com/aaroncbader/pystell_uw.git",
```
MicahGale, line 50: Some of these seem unnecessary additions from base conda, and can be removed. (openssl, etc.)

pyproject.toml
```diff
@@ -0,0 +1,68 @@
...
+]
+
+authors = [
+	{name = "Connor Moreno", email = "camoreno@wisc.edu"},
+	{name = "Edgar"}, # TODO
+	{name = "Paul Wilson", email = "paul.wilson@wisc.edu"},
+
+] # TODO
+
+keywords = [] # TODO
```
MicahGale, line 19: Need input @connoramoreno

pyproject.toml
```diff
@@ -0,0 +1,68 @@
...
+maintainers = [
+	{name = "Connor Moreno", email = "camoreno@wisc.edu"}
+]
+
+authors = [
+	{name = "Connor Moreno", email = "camoreno@wisc.edu"},
+	{name = "Edgar"}, # TODO
+	{name = "Paul Wilson", email = "paul.wilson@wisc.edu"},
+
+] # TODO
```
MicahGale, line 17: Should we add all contributors? I got a little lazy.

pyproject.toml
connoramoreno:
```suggestion
keywords = ["stellarator", "fusion", "reactor", "parametric", "cad", "model", "dagmc", "geometry"]
```

pyproject.toml
connoramoreno: Maybe we can chat about best practices here at Hacker Within

pyproject.toml
connoramoreno: To be chatted about at Hacker Within

pyproject.toml
```diff
@@ -0,0 +1,68 @@
...
+name = "parastell"
+description = "Parametric 3-D CAD modeling toolset for stellarator fusion devices; aka Micah pretends to know fusion"
+version = "0.0.2"
+
+readme = "README.md"
+requires-python = ">=3.11"
+maintainers = [
+	{name = "Connor Moreno", email = "camoreno@wisc.edu"}
+]
+
```
connoramoreno, line 11: What's the difference between "maintainers" and "authors"? Are the former those who actively maintain the software, whereas the latter are those who have contributed at some point?

pyproject.toml
```diff
@@ -0,0 +1,68 @@
+[project]
+name = "parastell"
+description = "Parametric 3-D CAD modeling toolset for stellarator fusion devices; aka Micah pretends to know fusion"
```
connoramoreno, line 3: 😉 
```suggestion
description = "Parametric 3-D CAD modeling toolset for stellarator fusion devices"
```

connoramoreno (changes requested):
Initial review to be expanded on at Hacker Within

MicahGale:
> Initial review to be expanded on at Hacker Within

I won't be there tomorrow FYI.

connoramoreno:
> > Initial review to be expanded on at Hacker Within
> 
> I won't be there tomorrow FYI.

Yeah, I provisionally cancelled it in favor of Stand Up for Science. Maybe we can just open the discussion in this PR

pyproject.toml
connoramoreno: @MicahGale, @gonuke, do you have any input on best practices when listing authors and maintainers?

pyproject.toml
connoramoreno: In the event that someone `pip` installs in a new (empty) environment, wouldn't those packages need to be installed?

pyproject.toml
MicahGale: My view is that anyone who has made a commit that is merged has the right to decide if they want to be listed as an author. For maintainers you should have < 3 generally. These are the people who are responding to issues. Part of this todo was laziness. 

Generally emails are a good idea. One caveat. It really messes up PyPI if you mix people with and without email addresses.

pyproject.toml
MicahGale: Yep. That's how I interpret it. There is apparently no strict guidance on this:



> Corresponding [core metadata](https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata) field: [Author](https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata-author), [Author-email](https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata-author-email), [Maintainer](https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata-maintainer), and [Maintainer-email](https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata-maintainer-email)

> The people or organizations considered to be the “authors” of the project. The exact meaning is open to interpretation — it may list the original or primary authors, current maintainers, or owners of the package.

> The “maintainers” key is similar to “authors” in that its exact meaning is open to interpretation.

> These keys accept an array of tables with 2 keys: name and email. Both values must be strings. The name value MUST be a valid email name (i.e. whatever can be put as a name, before an email, in [RFC 822](https://datatracker.ietf.org/doc/html/rfc822.html)) and not contain commas. The email value MUST be a valid email address. Both keys are optional, but at least one of the keys must be specified in the table.

-[PyPA](https://packaging.python.org/en/latest/specifications/pyproject-toml/#authors-maintainers)

MicahGale:
I'm ok with waiting a week if you are. I think this would allow good conversations about packaging.

pyproject.toml
MicahGale: Idea: Maintainers have emails; authors don't.

MicahGale: Add @gonuke as a maintainer.

connoramoreno: Decided on at Hacker Within: authors list includes all contributors, maintainers list are those designated for ParaStell correspondence

pyproject.toml
```diff
@@ -0,0 +1,68 @@
...
+       "Programming Language :: Python :: 3.12",
+       "Programming Language :: Python :: 3.13"
+]
+
+dependencies = [
+	"numpy==1.26.4",
+	"scipy",
+	"cadquery",
+	"moab=5.5.1[build=nompi_tempest_*]",
+	"cad_to_dagmc >=0.8.2",
```
gonuke, line 42: Is this space valid?

pyproject.toml
MicahGale: `pyhton ==3.11`

MicahGale: Cadquery has pip install issues.

For now: 

1. forsake pypi
2. Do conda-forge
3. Look into how to do `conda install` locally

MicahGale: Remove `ca-certificates`, `certifi` `openssl`.

PR pip installable `pystell_uw` and `pydagmc`

pyproject.toml
```diff
@@ -0,0 +1,68 @@
...
+	"openmc>=0.15.0[build=dagmc_nompi_*]",
+	"matplotlib",
+	"netcdf4",
+	"pyyaml",
+	"git+https://github.com/aaroncbader/pystell_uw.git",
+      	"git+https://github.com/svalinn/pydagmc.git"
+]
+
+[project.optional-dependencies]
+test = ["pytest"]
```
MicahGale, line 55: Could add `coveralls`

pyproject.toml
MicahGale: It actually is, but I will make it consistent.

parastell/source_mesh.py
```diff
@@ -602,7 +602,7 @@ def generate_source_mesh():
 
     source_mesh.export_mesh(
         export_dir=args.export_dir,
-        **(filter_kwargs(source_mesh_dict, ["filename"]))
+        **(filter_kwargs(source_mesh_dict, ["filename"])),
```
gonuke, line 605: Is this necessary? nice to have? does it add anything?

.github/workflows/ci.yml
```diff
@@ -10,9 +10,12 @@ on:
       - 'environment.yml'
 
 jobs:
-  test-dependency-img:
-    runs-on: ubuntu-latest
-    container: ghcr.io/svalinn/parastell-ci
+  runs-on: ubuntu-latest
+    test-dependency-img:
+      container: ghcr.io/svalinn/parastell-ci
```
gonuke, line 15: This change confuses me?

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
...
+      - run:  conda config --set pip_interop_enabled true
+        name: config pip-conda interop
+      - run: conda install -c conda-forge conda-build cadquery
+        name: Install conda dependencies
+      - run: conda develop .
+        name: Install locally with conda
+      - name: Install with pip
+        run: pip install .[develop]
+      - name: Test building wheels
+        run: python -m build --sdist --wheel
```
gonuke, line 34: This would be more readable with blank lines between steps and each step having a similar structure of `name` and then `run`

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
+name: Test packaging and installing
+
+on:
+  # allows us to run workflows manually
+  workflow_dispatch:
+  pull_request:
+    paths-ignore:
+      - 'Dockerfile'
+      - '.github/workflows/docker_publish.yml'
+      - 'environment.yml'
```
gonuke, line 10: Don't we rely on the environment for this build? Shouldn't be test this when the environment changes then?

gonuke (changes requested):
A few questions or comments on this draft

MicahGale:
I'm struggling with installing `cadquery` with conda and python 3.11. Have you seen this issue before @connoramoreno?

connoramoreno:
What's the issue you're having? My local environment uses Python 3.11.6 and CadQuery 2.5.2 (also what's usually installed via our `environment.yml`)

MicahGale:
Oh wait never mind that wasn't the issue. THe issue is that I'm trying to get it to install via `pip` locally and pip doesn't know cadquery is available. Ok I'll play around a bit more with that.

I would like to move away from needing an environment file.

MicahGale:
Here's a hacky work around for the time being.

MicahGale:
Goal for this PR:

1. Switch to `conda create ...`, `pip install .` workflow
2. Test it on GHA
3. Merge
4. Open `conda-forge` recipe PR.

.github/workflows/ci.yml
connoramoreno: We'll need to revert the first two lines
```suggestion
  test-dependency-img:
    runs-on: ubuntu-latest
    container: ghcr.io/svalinn/parastell-ci
```

.github/workflows/build.yml
MicahGale: I can't remember the resolution of this question @connoramoreno. Did we decide to not change this?

connoramoreno: I can't remember either but I think what Paul says makes sense. `environment.yml` should not be ignored since it defines our Python dependencies (and is referenced later in the workflow file).

parastell/source_mesh.py
MicahGale: I think this was just `black`.

.github/workflows/ci.yml
```diff
@@ -13,8 +13,12 @@ jobs:
   test-dependency-img:
     runs-on: ubuntu-latest
     container: ghcr.io/svalinn/parastell-ci
-
+    
     name: Perform CI Tests
+    strategy:
+      matrix:
+        python-version: ["3.11"]
```
gonuke, line 20: This doesn't appear to do anything... remove until it does?

gonuke (changes requested):
This seems ok to me, although I don't know much about the best practices for ’pyproject.toml` files

README.md
```diff
@@ -62,24 +62,18 @@ While it is possible to use ParaStell with older versions of Cubit, additional s
 
 If you do not have a Coreform Cubit license, you may be able to get one through [Cubit Learn](https://coreform.com/products/coreform-cubit/free-meshing-software/) at no cost.
 
-## Executing ParaStell Scripts with YAML Input
-While ParaStell can imported as a module to make use of its Python API, ParaStell also has an executable to alternatively call functionality via command line. This executable uses a YAML configuration file as a command-line argument to define input parameters.
+### Finally Install Parastell
 
-To make use of this feature, add the ParaStell `executables` directory to your `PATH` by adding a line similar to the following to your `.bashrc` file:
+Now that all dependencies have been installed you can install parastell with `pip`. Run the following command from the root of the parastell repository:
```
connoramoreno, line 67: Grammar, capitalization
```suggestion
Now that all dependencies have been installed, you can install ParaStell with `pip`. Run the following command from the root of the ParaStell repository:
```

examples/coils.example
connoramoreno, line 1: This is more of a comment for the `Examples` directory as a whole: was renaming the directory intentional? I'm not against it, just want to make sure.

examples/nwl_geom_example.py
connoramoreno, line 1: This file and `nwl_plot_example.py` were removed from the repository at some point, and should stay removed. I'm guessing something happened during your rebase.

examples/nwl_plot_example.py
connoramoreno, line 1: See comment on `nwl_geom_example.py`

parastell/__main__.py
connoramoreno, line 1: Is this file necessary? In other projects with conceptually similar packaging, I've only seen `__init__.py`

connoramoreno (changes requested):
I just have mostly high-level comments for now. Have you run the test suite locally after installing ParaStell to make sure everything works?

.github/workflows/ci.yml
MicahGale: The goal was to make adding 3.12, 3.13, etc. easy to add when they are supported. Though I need to make sure we actually use that infrastructure first...

examples/coils.example
MicahGale: Yes it was. `Examples` was confusing python, and generally is not pythonic.

parastell/__main__.py
MicahGale: Yes. [Here's the thorough documentation](https://docs.python.org/3/library/__main__.html). My understanding is that `__init__.py` is ran every time the package is imported. `__main__.py` is only executed when the package is called from the command line, e.g., `python -m parastell`. Yes this is duplication with the scripts being exposed, but I think this gives a good backup for when `$PATH` isn't cooperating. I should test this execution path though.

MicahGale:
> I just have mostly high-level comments for now. Have you run the test suite locally after installing ParaStell to make sure everything works?

I still need to debug what I broke with the CI tests.

.github/workflows/ci.yml
MicahGale: Ok upon further inspection, I was very naive. We would need to make multiple docker images, and test against those, so I removed this.

MicahGale:
Pin to cad_to_dagmc `0.9.0`

MicahGale:
TODO: move this to inside of `Svalinn` for CI to succeed.
* Open issue about not having access to CUBIT on CI.

PR #201, opened 2025-03-06
adds custom export directory to export_pydagmc_model method

### LINKED ISSUE #199: Custom export directory for pydagmc model
https://github.com/svalinn/parastell/blob/8e0d597516f290d3b80f1243ce897b3442170425/parastell/parastell.py#L588

Hi,

I am not sure if there is already a way to handle this, but currently this method saves the h5m dagmc file in the working directory. I was wondering if it would be possible to add an extra argument such as export_dir, where the export directory can be defined.
This is handled in other methods (e.g. export_invessel_build).
A workaround could be:
```python
from pathlib import Path

def export_pydagmc_model(self, filename="dagmc", export_dir=""):
    """Saves the PyDAGMC model to .h5m format.

    Arguments:
        filename (str): name of DAGMC output file, defaults to 'dagmc'
        export_dir (str): directory to save the output file, defaults to current directory
    """
    if not export_dir:
        export_dir = "."
    export_path = Path(export_dir) / Path(filename).with_suffix(".h5m")
    self.pydagmc_model.write_file(str(export_path))

tokamaster:
This PR adds the option to define a custom export directory for the dagmc model created with pydagmc.
Closes #199 .

### DIFF
--- a/parastell/parastell.py
+++ b/parastell/parastell.py
@@ -585,14 +585,16 @@ class Stellarator(object):
         else:
             self.pydagmc_model = self.invessel_build.dag_model
 
-    def export_pydagmc_model(self, filename="dagmc"):
+    def export_pydagmc_model(self, filename="dagmc", export_dir=""):
         """Saves the PyDAGMC model to .h5m format.
 
         Arguments:
             filename (str): name of DAGMC output file, defaults to 'dagmc'
+            export_dir (str): directory to save the output file, defaults to current directory
         """
-        filename = Path(filename).with_suffix(".h5m")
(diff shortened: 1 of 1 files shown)


connoramoreno:
Have you tested this addition to make sure it works?

tokamaster:
Yes, I have tried locally and it works as intended. It might be a good idea to add tests for this.

connoramoreno:
Perhaps the subject of a different PR 🙂

connoramoreno (approved):
LGTM, thanks @tokamaster

PR #203, opened 2025-03-19
Add method to convert DAGMC volume to STEP

### LINKED ISSUE #202: Method to Convert PyDAGMC in-vessel components to step format
While the PyDAGMC workflow works well for neutron transport, it would be good to have the ability to generate step files of components for use with other analysis tools, per @bonitagoh.

A previous attempt at the faceted geometry workflow used cubit to build the solids https://github.com/Edgar-21/parastell/tree/cubit_faceted_surfaces.

While this was limiting for a full in-vessel build I believe the approach will be useful for making step files out of single in-vessel components, using information contained in the `invessel_build.Surface` objects defining the inner and outer surface of the component.

I was thinking about adding a function to `cubit_utils` or `utils` that accepts two `Surface` objects (or just one to make a component with no thru hole) and creates a step file.

Edgar-21:
Closes #202 

Two methods are added to `parastell.utils`, one which reads STL files defining the surfaces of the volume to be converted and returns a CadQuery Solid defined by those surfaces, and another than streamlines the process by taking a PyDAGMC DAGModel object and a volume ID and writing the corresponding CadQuery Solid to STEP.

### DIFF
--- a/parastell/utils.py
+++ b/parastell/utils.py
@@ -321,7 +321,7 @@ def combine_dagmc_models(models_to_merge):
 
 
 def stl_surfaces_to_cq_solid(stl_file_paths, tolerance=0.001):
-    """ "Create a solid volume in CadQuery from STL files defining the surfaces
+    """Create a solid volume in CadQuery from STL files defining the surfaces
     of a watertight volume. The resulting volume should have surfaces that
     exactly match the STL files provided. Useful for obtaining CAD
     representations of volumes in DAGMC models.
@@ -334,7 +334,7 @@ def stl_surfaces_to_cq_solid(stl_file_paths, tolerance=0.001):
 
     Returns:
         cq_solid (CadQuery Solid): Solid volume with surfaces defined by the
-            the provided STL files.
+            the triangles in the provided STL files.
     """
     sewer = BRepBuilderAPI_Sewing(tolerance)
 
@@ -391,8 +391,9 @@ def dagmc_volume_to_step(
     if num_tris != num_faces:
         raise ValueError(
             f"Number of faces in the STEP file ({num_faces}) does"
-            "not match the number of triangles in the STL files"
-            f"({num_tris})"
+            "not match the number of triangles in the DAGMC volume"
+            f"({num_tris}). Please verify that your geometry is valid or use"
+            "a tighter tolerance."
         )
 
     cq_solid.exportStep(str(Path(step_file_path).with_suffix(".step")))
--- a/parastell/utils.py
+++ b/parastell/utils.py
@@ -2,12 +2,19 @@ import yaml
 import tempfile
 from pathlib import Path
 from functools import cached_property
+import tempfile
 
 import numpy as np
 import math
 from scipy.ndimage import gaussian_filter
 from pymoab import core, types
 import dagmc
+import cadquery as cq
+from OCP.BRepBuilderAPI import BRepBuilderAPI_Sewing
+from OCP.StlAPI import StlAPI_Reader
+from OCP.TopoDS import TopoDS_Shape
+from OCP.TopExp import TopExp_Explorer
+from OCP.TopAbs import TopAbs_SHELL
 
 
 m2cm = 100
@@ -233,7 +240,7 @@ class DAGMCRenumberizer(object):
     """Class to facilitate renumbering of entities to combine DAGMC models.
 
     Arguments:
-        mb (PyMOAB Core): PyMOAB core with a DAGMC models to be renumbered
+        mb (PyMOAB Core): PyMOAB core with the DAGMC models to be renumbered
         loaded. Optional.
     """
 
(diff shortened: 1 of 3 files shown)


parastell/utils.py
```diff
@@ -311,3 +318,82 @@ def combine_dagmc_models(models_to_merge):
...
+            volume to be converted to STEP.
+        volume_id (int): ID of the volume to be converted to STEP.
+        step_file_path (str): Path on which to save the resulting step file.
+        tolerance (float): Distance (in whatever units the STL file is in)
+            at which to consider vertices coincident or edges colinear.
+            Defaults to 0.001.
+    """
+    volume = dagmc_model.volumes_by_id[volume_id]
+
+    num_tris = len(volume.triangle_handles)
```
gonuke, line 375: Did you think about just writing these triangles (`volume.triangle_handles`) into a single STL? I have no idea whether it would work, just curious?

gonuke (commented):
This all looks pretty straightforward, but just one question out of ignorance that might simplify writing the STLs.

parastell/utils.py
Edgar-21: I'm not sure what it would look like to use the triangle handles directly - I know that both mbconvert and something like `vol.model.mb.write_file(stl_path, output_sets, sets=[vol.handle])` fail to write volumes to stl, while writing surfaces to stl works just fine. I also have had better luck making these volumes in CadQuery starting from surfaces, though that might be more related to user error when I was first figuring things out.

gonuke: It might be necessary to make a new set that contains the volume triangles - the write method may want a "set" rather than a "range" (which is what `volumes.triangle_handles` is)

gonuke: It may not be worth it - or we could make it an issue for an optional simplification in the future.

Edgar-21: Since I can't write the triangles in the volume directly to STL with the same method that works with surfaces, (seems to be some problem with MOAB) I think anything we come up with will be about the same level of complexity as writing each surface in a volume to STL. I'm inclined to leave it as is.

pshriwise: Sorry to interject out of the blue, can't seem to help it.

Is this perhaps because the volume set doesn't contain any triangles in the MOAB data model? (i.e. in the call to `write_file` below pass all surface handles of the volume instead of the volume handle itself)

Apologies if this ends up being a distraction.

Edgar-21: I was thinking of bugging you about this anyway, glad you showed up!

In PyDAGMC, there is a [method to write to VTK](https://github.com/svalinn/pydagmc/blob/c6621b77428b2ebb04c5cd6b7a8837655ab35800/dagmc/dagnav.py#L291) and below for reference.

```    
def to_vtk(self, filename):
        """Write the set to a VTK file. This will recursively gather all triangles under
        the group, volume or surface and generate a VTK file.
        """
        if not filename.endswith('.vtk'):
            filename += '.vtk'
        self.model.mb.write_file(filename, output_sets=[self.handle])

```
This can successfully write all the triangles in a volume to VTK, but if I try to do the same operation but change the suffix to `STL` it fails for volumes, while it succeeds for surfaces. I've noticed the same behavior with mbconvert, where I can write whole volumes to vtk but not stl, while surfaces can be written to either. The (seemingly) same issue showing up in both PyMOAB and mbconvert is what makes me suspicious that there is something about the STL writer that doesn't like writing volumes. I don't really have a strong enough handle on MOAB to understand why. Maybe its something to do with there being more than one set of triangles in a volume, but only a single set in a surface?

gonuke: I'm surprised that the VTK writer works on volume sets.  The `pymoab.write_file()` method expects the `ouput_sets` to be either an entity set or a range of *only* entity sets.  As @pshriwise mentions, volumes are entity sets, but don't contain anything in their sets - they only have parent-child relationships with their surfaces.  This is obscured somewhat in the pydagmc interface since there is a method `triangle_handles` that will loop over all the surfaces and collect their triangles into a range.

Therefore, I think this will not work:
`self.model.mb.write_File(filename, output_sets=[volume.handle])`
but this will:
`self.model.mb.write_File(filename, output_sets=volume.triangle_handles)` (I don't think this range needs to be passed as a list????)

pshriwise: Right, I think the issue is that the volume entity set itself doesn't contain any triangles. See the image on the data model below from one of @gonuke's slide decks.

image


The VTK writer in MOAB seems to traverse any child sets of the `output_sets` argument to collect elements -- probably because the VTK format is more generic, accommodating many different element types/dimensions so it does more work. In contrast, STL files (generally) contain single groups of triangles (this is how an STL will be read into MOAB, for instance, so I think MOAB is expecting to find the elements in the set(s) passed to the `write_file` call. This doesn't mean the STL writer in MOAB _shouldn't_ be traversing the child sets like the VTK reader, just an explanation as to why it may not based on the allowed contents of these two different file types.

So, rather than passing the volume handle, collecting the handles of the volume's surface sets (which _do_ contain triangle elements) would do the trick. It would look something like:

```python
model.mb.write_file('vol.stl', output_sets=[s.handle for s in self.surfaces])
```

I gave this a try using one of the `.h5m` files in the OpenMC DAGMC regression tests and it seemed to work alright.

Edgar-21: Here's a simple example:

```
import dagmc
dm = dagmc.DAGModel('dagmc.h5m')
vol = dm.volumes_by_id[2]
print('writing to vtk...')
vol.model.mb.write_file('vol2.vtk',output_sets=[vol.handle])
print('writing to stl...')
vol.model.mb.write_file('vol2.stl',output_sets=[vol.handle])
```
and the output
```
writing to vtk...
writing to stl...
Writer with name STL for file vol2.stl using extension stl (file type "(null)") was unsuccessful
MOAB ERROR: --------------------- Error Message ------------------------------------
MOAB ERROR: No triangles to write!
MOAB ERROR: write_file() line 96 in src/io/WriteSTL.cpp
Using default writer WriteHDF5 for file vol2.stl 
```
The vtk file is valid, I can open it up in paraview just fine. It does write a file for the stl, but writes to HDF5 with an STL suffix.

Attempting to write the triangle handles with:
```
vol.model.mb.write_file('vol2.stl', output_sets=vol.triangle_handles)
```
give this error
```
Cell In[2], line 1
----> 1 vol.model.mb.write_file('vol2.stl', output_sets=vol.triangle_handles)

File pymoab/core.pyx:153, in pymoab.core.Core.write_file()

OSError: Only EntitySets should be passed to write file.
```

pshriwise: Whoops, didn't see your comment here @gonuke. I think ``output_sets` does need to be either an `EntitySet` or Iterable of `EntitySet`'s, but it's been a while and I could be wrong here.

Edgar-21: Ah I got it `vol.model.mb.write_file('vol2.stl', output_sets=vol._get_triangle_sets())` works. `output_sets` needs a range of `EntitySet`

Edgar-21: I need to refresh PR pages more often I missed all the comments in the past half hour and was just blissfully writing my own :)

parastell/utils.py
```diff
@@ -311,3 +318,74 @@ def combine_dagmc_models(models_to_merge):
...
+
+    with tempfile.NamedTemporaryFile(delete=True, suffix=".stl") as temp_file:
+        stl_path = temp_file.name
+        volume.model.mb.write_file(
+            stl_path, output_sets=volume._get_triangle_sets()
+        )
+        cq_solid = stl_to_cq_solid(stl_path, tolerance)
+    num_faces = len(cq_solid.Faces())
```
gonuke, line 381: Since it is only used here, can we move the definition of `num_tris` to here?

tests/test_utils.py
```diff
@@ -81,3 +81,31 @@ def test_expand_list():
...
+    """
+    vol_id = 1
+    dag_model = dagmc.DAGModel("files_for_tests/one_cube.h5m")
+    vol = dag_model.volumes_by_id[vol_id]
+    dagmc_volume_volume = vol.volume
+    num_tris = len(vol.triangle_handles)
+
+    with tempfile.NamedTemporaryFile(delete=True, suffix=".stl") as temp_file:
+        stl_path = temp_file.name
+        vol.model.mb.write_file(stl_path, output_sets=vol._get_triangle_sets())
```
gonuke, line 104: as above
```suggestion
        dagmc_model.mb.write_file(stl_path, output_sets=[s.handle for s in vol.surfaces])
```

parastell/utils.py
```diff
@@ -311,3 +318,74 @@ def combine_dagmc_models(models_to_merge):
...
+            at which to consider vertices coincident or edges colinear.
+            Defaults to 0.001.
+    """
+    volume = dagmc_model.volumes_by_id[volume_id]
+
+    num_tris = len(volume.triangle_handles)
+
+    with tempfile.NamedTemporaryFile(delete=True, suffix=".stl") as temp_file:
+        stl_path = temp_file.name
+        volume.model.mb.write_file(
```
gonuke, line 377: Since you already have the model, this may be better
```suggestion
        dagmc_model.mb.write_file(
```

tests/test_utils.py
```diff
@@ -81,3 +81,31 @@ def test_expand_list():
...
+        cq_solid = stl_to_cq_solid(stl_path)
+
+    cq_solid_volume = cq_solid.Volume()
+    num_faces = len(cq_solid.Faces())
+
+    assert num_faces == num_tris
+    assert math.isclose(dagmc_volume_volume, cq_solid_volume)
```
gonuke, line 111: either pytest or numpy have a way to do this instead of `math`?

tests/test_utils.py
```diff
@@ -81,3 +81,31 @@ def test_expand_list():
...
+      * the number of triangle handles belonging to the DAGMC volume is the
+        same as the number of faces belonging to the CadQuery solid.
+      * the volume of the DAGMC volume is close to the volume of the CadQuery
+        solid (using math.isclose()).
+    """
+    vol_id = 1
+    dag_model = dagmc.DAGModel("files_for_tests/one_cube.h5m")
+    vol = dag_model.volumes_by_id[vol_id]
+    dagmc_volume_volume = vol.volume
+    num_tris = len(vol.triangle_handles)
```
gonuke, line 100: Move this close to the definition of `num_faces` since they are used together

gonuke (changes requested):
I think the simplification was worth it - now a few final touch ups.

gonuke (approved):
Thanks for this addition @Edgar-21

PR #204, opened 2025-03-22
Fix error for edge case when computing magnet coil outboard mid-plane index

connoramoreno:
A small modification to fix an edge-case error when computing the outboard mid-plane index for magnet coils. The edge-case occurred when the outboard mid-plane point was at z = 0. For such a point, the mid-plane flag, which indicates whether the point is one of two mid-plane points, is computed as 0 due to the nature of the `np.sign` function. The mid-plane flags are computed using a comparison of z-coordinate signs between adjacent filament points. Mid-plane flag values are supposed to be 1 for mid-plane point and -1 for all others, but `np.sign(0) = 0`. The solution here is to perturb the z-coordinates of all points (within the method) by a small amount `eps = 1e-10`.


gonuke:
Isn't there the risk that you perturbed the points so that the new point is exactly 0?  I wonder if there is a more robust logical approach?

gonuke:
E.g could you just test to see if any of the values are exactly 0 before you do the `np.sign` test?

connoramoreno:
There is that risk but it feels unlikely. A more robust alternative could make use of the [`np.place`](https://numpy.org/doc/2.1/reference/generated/numpy.place.html) function, replacing 0 values with `eps`.

connoramoreno:
Alternatively, we could just use a logical workflow to compare adjacent z-values; I was trying to avoid conditionals for computational expediency but maybe the difference wouldn't be too significant.

parastell/magnet_coils.py
```diff
@@ -440,12 +440,21 @@ def get_ob_mp_index(self):
...
             outboard_index (int): index of the outboard midplane point.
         """
+        # Define small value
+        eps = 1e-10
+        # Shift some coordinates by eps to compute appropriate midplane flags
+        # If z = 0 => midplane flag = 0 => incorrect maximum radius computed
+        # => incorrect OB midplane index
+        # Replace z-coordinates at 0 with eps
+        coords = self.coords
+        np.place(coords[:, 2], coords[:, 2] == 0.0, [eps])
```
gonuke, line 450: What about this?
```suggestion
        np.place(coords[:, 2], np.abs(coords[:, 2]) < eps, [eps])
```

gonuke (changes requested):
One suggestion to avoid a strict comparison with 0.0 - although I know this is actually the case we care about :)

gonuke:
Do you have a test for this so we can see that this has fixed it?

connoramoreno:
I've added a test in the single filament block and modified the coils example test file to reflect our edge case

tests/files_for_tests/coils.example
```diff
@@ -258,5 +258,5 @@ mirror NIL
   1.237565317338557E+01  -1.018174898724600E+01   1.933202743533675E-01   9.825067085820705E+06
   1.239173356270819E+01  -1.041975174744141E+01   5.196839261284493E-02   9.825067085820705E+06
   1.236279152161795E+01  -1.065567485505797E+01  -8.915871208808870E-02   9.825067085820705E+06
-  1.228696590129009E+01  -1.088179419180126E+01  -2.283491710711771E-01   0.000000000000000E+00 10 RotatedCurve5
+  1.228696590129009E+01  -1.088179419180126E+01   0.000000000000000E+00   0.000000000000000E+00 10 RotatedCurve5
```
gonuke, line 261: This point seems to be further away from z=0 than the preceding point - it may not affect the test, but it may be an invalid set of points???

gonuke (changes requested):
Just a question about whether your manual change is valid - it might not matter here, but worried about future implications of invalid coil data?

tests/files_for_tests/coils.example
connoramoreno: That's a good catch. I forgot that this file actually has two filaments in it and I modified the first coordinate in the first filament and the final coordinate in the other filament

connoramoreno:
We're still facing a similar issue though. Let me think about how to modify this.

connoramoreno:
I've removed a point from the filament just preceding the final point. Should give a valid coil set.

gonuke (approved):
Thanks @connoramoreno

tests/test_magnet_coils.py
```diff
@@ -82,12 +82,14 @@ def test_single_filament(single_filament):
...
     com_exp = np.array([66.66666667, 100.0, 33.33333333])
     com_toroidal_angle_exp = 0.982793723247329
+    ob_mp_idx_exp = 0
 
     assert np.allclose(tangents_exp, single_filament.tangents)
     assert np.allclose(com_exp, single_filament.com)
     assert np.isclose(
         com_toroidal_angle_exp, single_filament.com_toroidal_angle
     )
+    assert single_filament.get_ob_mp_index() == ob_mp_idx_exp
```
gonuke, line 92: The `single_filament` doesn't appear to actually cross the z-axis.  Is this a good test data set at all for this?

gonuke (changes requested):
I'm confused about whether our test for this is robust enough

gonuke (changes requested):
This looks better now.  I think we have shown that we don't necessarily test all our coil algorithms across enough different edge cases.  I don't think we need to brainstorm all those edge cases now, but it might be good to have two different single filament sets that we test, one that would have worked before this change and one that would have failed before this change.  I think the updated test filament would have always failed before this change.

gonuke (approved):
LGTM - thanks @connoramoreno

PR #208, opened 2025-03-25
Fix bug where zero volume coils make the material tag list too long in CAD-to-DAGMC workflow

### LINKED ISSUE #207: Mismatch between number of tags and number of triangle sets CAD-to-DAGMC
@fnovais11 noticed a bug with the CAD-to-DAGMC workflow where occasionally there would be more material tags than triangle sets, resulting in an error.

I believe this is due to our conservative means of estimating if a coil solid will be within the toroidal extent specified for the coils. After cutting to the toroidal extent specified, we completely remove some coil solids, but still pass material tags for them.

Edgar-21:
closes #207 

This PR introduces a check that only adds coil solids to `MagnetSetFromFilaments.coil_solids` if it has volume > 0. Zero volume solids can occur if a coil is completely outside of the toroidal extent, which can happen due to our conservative means of estimating if a coil will lie in the toroidal extent after building the coil solid.

Also noticed a comment that was a little too long and took care of that.


gonuke (changes requested):
Seems reasonable to me.  Do we have a test case that will show the failure and then that this fixes it?

parastell/magnet_coils.py
```diff
@@ -355,7 +355,11 @@ def build_magnet_coils(self):
 
         self._cut_magnets()
 
-        self.coil_solids = [coil.solid for coil in self.magnet_coils]
+        self.coil_solids = [
+            coil.solid
+            for coil in self.magnet_coils
+            if coil.solid.Volume() != 0
+        ]
```
connoramoreno, line 362:
```suggestion
        self.magnet_coils = [coil for coil in self.magnet_coils if coil.solid.Volume() != 0]
        
        self.coil_solids = [coil.solid for coil in self.magnet_coils]
```

parastell/parastell.py
```diff
@@ -590,7 +590,8 @@ def export_pydagmc_model(self, filename="dagmc", export_dir=""):
 
         Arguments:
             filename (str): name of DAGMC output file, defaults to 'dagmc'
-            export_dir (str): directory to save the output file, defaults to current directory
+            export_dir (str): directory to save the output file, defaults to
+            current directory
```
connoramoreno, line 594:
```suggestion
            export_dir (str): directory to save the output file, defaults to
                current directory
```

connoramoreno (approved):
After formatting for suggested commit is fixed, I'm happy to merge. Thanks @Edgar-21!

PR #209, opened 2025-04-01
Update NWL utility

### LINKED ISSUE #205: Relax root finding tolerance for NWL serach
The NWL utility currently uses root finding to map (x,y,z) coordinates onto (s, phi, theta) coordinates on the first wall based on the VMEC representation of the wall.  This root finding is extremely slow.  There are a variety of things we can do to speed this upe.

One option is to use a very loose tolerance on the root finding procedure: since we are seeking (s, phi, theta) only for the purpose of binning the result, we really only need to know the values within the resolution of our (phi, theta) grid.  This is generally quite low fidelity, so we should be able to use a large convergence tolerance on the root finding.

connoramoreno:
Updates the NWL utility, making modifications to its namespace, code structure, and documentation, as well as combining the two NWL example scripts into one and introducing testing.

Furthermore, enables relaxation of the root-finding algorithm, introducing a tolerance parameter. This parameter corresponds to `vol_tol` in [SciPy's DIRECT algorithm](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.direct.html). Relaxing this tolerance from SciPy's default of `1e-16` to ParaStell's now-default `1e-6` corresponds to roughly half an order of magnitude decrease in computation time, with no discernible sacrifice in accuracy.

The longer-term goal for the NWL utility is to switch from a root-finding routine to solve for surface crossing poloidal angles to some kind of search routine, for computational expediency. I was planning on delegating that task to a separate PR, introducing the tolerance as a quick, short-term solution while focusing on a broader update for the utility, but we can make that modification here too if desired.

Closes #205


parastell/nwl_utils.py
```diff
@@ -95,118 +87,129 @@ def nwl_transport(dagmc_geom, source_mesh, tor_ext, ss_file, num_parts):
...
+        poloidal_angles (list): poloidal angles of surface crossings [rad].
     """
+    toroidal_angles = np.arctan2(coords[:, 1], coords[:, 0])
+    chunk_size = math.ceil(len(toroidal_angles) / num_threads)
 
-    phi_coords = np.arctan2(coords[:, 1], coords[:, 0])
-    chunk_size = math.ceil(len(phi_coords) / num_threads)
     chunks = []
+
     for i in range(num_threads):
```
gonuke, line 174:
```suggestion
    for chunk_start in range(0, len(phi_coords), chunk_size):
```

parastell/nwl_utils.py
```diff
@@ -95,118 +87,129 @@ def nwl_transport(dagmc_geom, source_mesh, tor_ext, ss_file, num_parts):
...
-        chunk = list(
-            zip(
-                phi_coords[i * chunk_size : (i + 1) * chunk_size],
+        chunks.append(
+            (
+                vmec_file,
+                wall_s,
+                conv_tol,
+                toroidal_angles[i * chunk_size : (i + 1) * chunk_size],
                 coords[i * chunk_size : (i + 1) * chunk_size],
```
gonuke, line 181:
```suggestion
                toroidal_angles[chunk_start: chunk_start + chunk_size],
                coords[chunk_start: chunk_start + chunk_size],
```

parastell/nwl_utils.py
```diff
@@ -215,192 +218,213 @@ def extract_coords(source_file):
...
+    toroidal_angles = []
+    poloidal_angles = []
 
-    phi_coords = []
-    theta_coords = []
+    batch_size = math.ceil(len(coords) / num_batches)
 
-    if chunk_size is None:
-        chunk_size = len(coords)
+    for i in range(num_batches):
```
gonuke, line 310:
```suggestion
    for batch_start range(0, len(coords), batch_size):
```

parastell/nwl_utils.py
```diff
@@ -215,192 +218,213 @@ def extract_coords(source_file):
...
-    chunks = math.ceil(len(coords) / chunk_size)
-
-    for i in range(chunks):
-        phi_coord_subset, theta_coord_subset = flux_coords(
-            plas_eq,
+        toroidal_angle_batch, poloidal_angle_batch = compute_flux_coordinates(
+            vmec_file,
             wall_s,
-            coords[i * chunk_size : (i + 1) * chunk_size],
+            coords[i * batch_size : (i + 1) * batch_size],
```
gonuke, line 316:
```suggestion
            coords[batch_start : batch_start + batch_size],
```

parastell/nwl_utils.py
```diff
@@ -215,192 +218,213 @@ def extract_coords(source_file):
...
+
+    # Compute centroids of bins
+    toroidal_centroids = np.linspace(
+        0.0, toroidal_extent, num=num_toroidal_bins
+    )
+    poloidal_centroids = np.linspace(
+        0.0, poloidal_extent, num=num_poloidal_bins
+    )
+
+    num_particles = len(coords)
```
gonuke, line 359: Move this definition earlier since you use `len(coords)` in a few places

parastell/nwl_utils.py
```diff
@@ -215,192 +218,213 @@ def extract_coords(source_file):
...
+    conv_tol=1e-6,
+    num_batches=1,
     num_crossings=None,
-    chunk_size=None,
-    num_threads=1,
+    num_threads=None,
+    logger=None,
 ):
-    """Computes and plots NWL. Assumes toroidal extent is less than 360 degrees
+    """Computes and plots NWL. Assumes toroidal extent is less than 360 degrees.
```
gonuke, line 262: Plotting now happens elsewhere
```suggestion
    """Computes NWL. Assumes toroidal extent is less than 360 degrees.
```

gonuke (changes requested):
This was surprisingly hard to review, given that it was mostly variable name changes...  I think it would have been better to do it on a function-by-function basis for rapid review.

A few minor changes in indexing variable for readability.

parastell/nwl_utils.py
```diff
@@ -215,192 +218,217 @@ def extract_coords(source_file):
...
+    batch_size = math.ceil(num_particles / num_batches)
 
-    chunks = math.ceil(len(coords) / chunk_size)
+    iterator = 1
 
-    for i in range(chunks):
-        phi_coord_subset, theta_coord_subset = flux_coords(
-            plas_eq,
+    for batch_start in range(0, num_particles, batch_size):
+        iterator += 1
```
gonuke, line 315: You don't need `iterator` if you use

```suggestion
        batch_num = batch_start/batch_size + 1
```

gonuke (changes requested):
One last change on this iterator

parastell/nwl_utils.py
connoramoreno: I was thinking of doing something like that but I figured if we're going to save the value anyways, something like `iterator += 1` is more readable? Although in either case, `batch_num` is probably the better variable name to use. What do you think?

gonuke: If you want to use a variable that iterates like this, then you can/should use `enumerate`

gonuke (approved):
This looks good to me - thanks @connoramoreno

PR #210, opened 2025-04-04
Add open-source volumetric meshing routines

### LINKED ISSUE #140: Provide open-source volumetric meshing of magnets and in-vessel components
To provide a completely open-source, license-free workflow, it will be necessary to add an alternative volumetric meshing scheme for the magnets and in-vessel components.

connoramoreno:
This branch introduces open-source tetrahedral meshing workflows for the `InVesselBuild` and `MagnetSet` classes.

The `InVesselBuild` class has two meshing routines added, one using MOAB and another using Gmsh.

- The MOAB routine, analogous to the PyDAGMC workflow, uses the IVB point cloud to generate a mesh in MOAB for a single specified component. In the resulting mesh, the component is one tetrahedron thick. In fact, one purpose of this MOAB workflow is to ensure a one-tetrahedron-thick mesh (useful for mesh tallies). This capability may be expanded to include multiple components in a single mesh.
- The Gmsh routine is usable with both CadQuery and PyDAGMC geometries. When used in conjunction with CadQuery geometries, CadQuery CAD solids are directly imported and meshed. When used with PyDAGMC geometries, PyDAGMC volumes are imported and remeshed according to specified parameters.

The `MagnetSet` class has one meshing routine added, using Gmsh. Like the `InVesselBuild` Gmsh routine, this routine directly imports CadQuery CAD solids and meshes those.

List of changes:

- Adds MOAB and Gmsh meshing routines for the `InVesselBuild` class
- Adds Gmsh meshing routine for the `MagnetSet` class
- Adds tests to cover new functionality
- Standardizes Cubit meshing methods across ParaStell classes
- Standardizes export method naming convention
- Updates tests and examples to reflect changes in API
- Introduces `ToroidalMesh` abstract base class

Closes #140

### DIFF
--- a/parastell/invessel_build.py
+++ b/parastell/invessel_build.py
@@ -710,25 +710,16 @@ class InVesselBuild(object):
 
         return solids, mat_tags
 
-    def mesh_components_moab(
-        self, component, remesh=False, min_mesh_size=5.0, max_mesh_size=20.0
-    ):
-        """Creates a tetrahedral mesh of in-vessel component volumes via MOAB
-        and, optionally, Gmsh. If the initial MOAB mesh, created using the
-        point clouds of the specified components, is not subsequently remeshed
-        in Gmsh, each component in the mesh will be one tetrahedron thick.
+    def mesh_components_moab(self, component):
+        """Creates a tetrahedral mesh of a single in-vessel component volume
+        via MOAB. This mesh is created using the point cloud of the specified
+        component and as such, the mesh will be one tetrahedron thick.
 
         Arguments:
-            component (str): name of the in-vessel component to be meshed.
-            remesh (bool): remesh/refine the initial MOAB mesh in Gmsh
-                (defaults to False).
-            min_mesh_size (float): minimum size of mesh elements for a Gmsh
-                remesh (defaults to 5.0).
-            max_mesh_size (float): maximum size of mesh elements for a Gmsh
-                remesh (defaults to 20.0).
+            component (str): name of in-vessel component to be meshed.
         """
         self._logger.info(
-            "Generating tetrahedral mesh of in-vessel component(s) via MOAB..."
+            f"Generating tetrahedral mesh of {component} volume via MOAB..."
         )
 
         self.mesh_mbc = core.Core()
@@ -737,6 +728,8 @@ class InVesselBuild(object):
         surface_keys = list(self.Surfaces.keys())
 
         for i, key in enumerate(surface_keys[:-1]):
+            # Inner surface of given component = outer surface of inner
+            # component
             if surface_keys[i + 1] == component:
                 surfaces.append(self.Surfaces[key])
         surfaces.append(self.Surfaces[component])
@@ -764,7 +757,7 @@ class InVesselBuild(object):
             poloidal_idx (int): index defining location along poloidal angle axis.
             toroidal_idx (int): index defining location along toroidal angle axis.
         """
-        # relative offsets of vertices in a 3-D index space
+        # Relative offsets of vertices in a 3-D index space
         hex_vertex_stencil = np.array(
             [
                 [0, 0, 0],
@@ -778,7 +771,7 @@ class InVesselBuild(object):
             ]
         )
 
-        # Ids of hex vertices applying offset stencil to current point
+        # IDs of hex vertices applying offset stencil to current point
         hex_idx_data = (
             np.array([0, poloidal_idx, toroidal_idx]) + hex_vertex_stencil
         )
@@ -790,9 +783,9 @@ class InVesselBuild(object):
         # Define MOAB canonical ordering of hexahedron vertex indices
         # Ordering follows right hand rule such that the fingers curl around
         # one side of the tetrahedron and the thumb points to the remaining
-        # vertex. The vertices are ordered such that those on the side are
-        # first, ordered clockwise relative to the thumb, followed by the
-        # remaining vertex at the end of the thumb.
+        # vertex. The vertices are ordered such that those on a face are first,
+        # ordered clockwise relative to the thumb, followed by the remaining
+        # vertex at the end of the thumb.
         # See Moreno, Bader, Wilson 2024 for hexahedron splitting
         hex_to_tets_mapping = [
             [idx_list[0], idx_list[3], idx_list[1], idx_list[4]],
@@ -811,8 +804,8 @@ class InVesselBuild(object):
         (Internal function not intended to be called externally)
 
         Arguments:
-            vert_idx (list of int): list of vertex
-                [flux surface index, poloidal angle index, toroidal angle index]
+            vertex_idx (list): vertex's 3-D grid indices in order
+                [surface index, poloidal angle index, toroidal angle index]
 
         Returns:
             id (int): vertex index in row-major order as stored by MOAB
@@ -853,6 +846,10 @@ class InVesselBuild(object):
             export_dir (str): directory to which to export the h5m output file
                 (defaults to empty string).
         """
+        self._logger.info(
+            "Exporting mesh H5M file for in-vessel component(s)..."
+        )
+
         export_path = Path(export_dir) / Path(filename).with_suffix(".h5m")
         self.mesh_mbc.write_file(str(export_path))
 
--- a/parastell/invessel_build.py
+++ b/parastell/invessel_build.py
@@ -731,7 +731,7 @@ class InVesselBuild(object):
         surfaces = [self.Surfaces[surface_keys[component_idx - 1]]]
         surfaces.append(self.Surfaces[component])
 
-        self.moab_mesh = InVesselComponentMesh(surfaces)
+        self.moab_mesh = InVesselComponentMesh(surfaces, self._logger)
 
         self.moab_mesh.create_vertices()
         self.moab_mesh.create_mesh()
@@ -745,7 +745,6 @@ class InVesselBuild(object):
             export_dir (str): directory to which to export the h5m output file
                 (defaults to empty string).
         """
-        self._logger.info("Exporting mesh H5M file for in-vessel component...")
         self.moab_mesh.export_mesh(filename, export_dir=export_dir)
 
     def mesh_components_gmsh(
@@ -863,9 +862,7 @@ class InVesselBuild(object):
             export_dir (str): directory to which to export the h5m output file
                 (defaults to empty string).
         """
-        self._logger.info(
-            "Exporting mesh H5M file for in-vessel component(s)..."
-        )
+        self._logger.info("Exporting mesh H5M file...")
 
         vtk_path = Path(export_dir) / Path(filename).with_suffix(".vtk")
         moab_path = vtk_path.with_suffix(".h5m")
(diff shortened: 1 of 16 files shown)


parastell/magnet_coils.py
```diff
@@ -88,19 +94,69 @@ def mesh_magnets(self, min_size=20.0, max_size=50.0, max_gradient=1.5):
...
+
+        for solid in self.coil_solids:
+            gmsh.model.occ.importShapesNativePointer(solid.wrapped._address())
+
+        gmsh.model.occ.synchronize()
+
+        gmsh.option.setNumber("Mesh.MeshSizeMin", min_mesh_size)
+        gmsh.option.setNumber("Mesh.MeshSizeMax", max_mesh_size)
+
+        gmsh.model.mesh.generate(3)
```
gonuke, line 136: magic number: 3??

parastell/magnet_coils.py
```diff
@@ -88,19 +94,69 @@ def mesh_magnets(self, min_size=20.0, max_size=50.0, max_gradient=1.5):
...
+        Arguments:
+            filename (str): name of H5M output file (defaults to
+                'magnet_mesh').
+            export_dir (str): directory to which to export the h5m output file
+                (defaults to empty string).
+        """
+        self._logger.info("Exporting mesh H5M file for magnets...")
+
+        vtk_path = str(Path(export_dir) / Path(filename).with_suffix(".vtk"))
+        moab_path = Path(export_dir) / Path(filename).with_suffix(".h5m")
```
gonuke, line 151:
```suggestion
        moab_path = vtk_path.with_suffix(".h5m")
```

parastell/invessel_build.py
```diff
@@ -199,6 +202,10 @@ def populate_surfaces(self):
             expand_list(self.radial_build.poloidal_angles, self.num_rib_pts)
         )
 
+        # Store actual number of ribs and rib points
+        self._num_ribs = len(self._toroidal_angles_exp)
+        self._num_rib_pts = len(self._poloidal_angles_exp)
```
gonuke, line 207: How do these relate to `self.num_ribs` and `self.num_rib_pts`?  I' concerned that we have both defined and they could be different?

parastell/invessel_build.py
```diff
@@ -501,32 +508,325 @@ def extract_solids_and_mat_tags(self):
...
+
+        vtk_path = str(Path(export_dir) / Path(filename).with_suffix(".vtk"))
+        moab_path = Path(export_dir) / Path(filename).with_suffix(".h5m")
+
+        gmsh.write(vtk_path)
+
+        gmsh.clear()
+        gmsh.finalize()
+
+        os.system(f"mbconvert {vtk_path} {moab_path}")
```
gonuke, line 781: can't we use pymoab to do this? I have a vague recollection that there was some reason we couldn't but not sure?

parastell/invessel_build.py
```diff
@@ -501,32 +508,325 @@ def extract_solids_and_mat_tags(self):
...
+        self.mesh_mbc = core.Core()
+
+        surfaces = []
+        surface_keys = list(self.Surfaces.keys())
+
+        for i, key in enumerate(surface_keys[:-1]):
+            # Inner surface of given component = outer surface of inner
+            # component
+            if surface_keys[i + 1] == component:
+                surfaces.append(self.Surfaces[key])
```
gonuke, line 532: rely 
```suggestion
        component_index = surface_keys.index(component)
        surfaces = [self.Surfaces(surface_keys[component_index - 1])]
```

parastell/invessel_build.py
```diff
@@ -501,32 +508,325 @@ def extract_solids_and_mat_tags(self):
...
+        self.verts = self.mesh_mbc.create_vertices(coords)
+
+        self.mesh_set = self.mesh_mbc.create_meshset()
+        self.mesh_mbc.add_entity(self.mesh_set, self.verts)
+
+        for toroidal_idx in range(self._num_ribs - 1):
+            for poloidal_idx in range(self._num_rib_pts - 1):
+                self._create_tets_from_hex(poloidal_idx, toroidal_idx)
+
+    def _create_tets_from_hex(self, poloidal_idx, toroidal_idx):
```
gonuke, line 550: Is this different from the one in our source mesh code? Can't we reuse that?

parastell/invessel_build.py
```diff
@@ -501,32 +508,325 @@ def extract_solids_and_mat_tags(self):
...
+            [idx_list[1], idx_list[3], idx_list[2], idx_list[6]],
+            [idx_list[1], idx_list[4], idx_list[6], idx_list[5]],
+            [idx_list[3], idx_list[6], idx_list[4], idx_list[7]],
+            [idx_list[1], idx_list[3], idx_list[6], idx_list[4]],
+        ]
+
+        for vertex_ids in hex_to_tets_mapping:
+            self._create_tet(vertex_ids)
+
+    def _get_vertex_id(self, vertex_idx):
```
gonuke, line 599: we should reuse code from source mesh code

parastell/invessel_build.py
```diff
@@ -501,32 +508,325 @@ def extract_solids_and_mat_tags(self):
...
+        gmsh.model.mesh.classifySurfaces(
+            np.deg2rad(angle),
+            includeBoundary,
+            forceParameterizablePatches,
+            np.deg2rad(curveAngle),
+        )
+
+        gmsh.model.mesh.create_geometry()
+
+        surfaces = gmsh.model.getEntities(2)
```
gonuke, line 738: magic number: 2

parastell/invessel_build.py
```diff
@@ -501,32 +508,325 @@ def extract_solids_and_mat_tags(self):
...
+            filename (str): name of H5M output file.
+            export_dir (str): directory to which to export the h5m output file
+                (defaults to empty string).
+        """
+        self._logger.info(
+            "Exporting mesh H5M file for in-vessel component(s)..."
+        )
+
+        vtk_path = str(Path(export_dir) / Path(filename).with_suffix(".vtk"))
+        moab_path = Path(export_dir) / Path(filename).with_suffix(".h5m")
```
gonuke, line 774:
```suggestion
        moab_path = vtk_path.with_suffix(".h5m")
```

parastell/parastell.py
```diff
@@ -178,26 +178,71 @@ def export_invessel_build(self, export_dir=""):
         """
         self.invessel_build.export_step(export_dir=export_dir)
 
-    def export_invessel_component_mesh(
-        self, components, mesh_size=5, import_dir="", export_dir=""
+    def export_invessel_build_mesh_moab(
```
gonuke, line 181: This one only works on a single component while the others seem to work on multiple components.... can we make them more similar?

gonuke (changes requested):
There is a lot of repeated code used for the MOAB based meshing - probably good to find a way to resuse the code from `SourceMesh`. 

One option is to make a parent class called `LayerMesh` that handles all the meshing stuff, and have `SourceMesh` derive from that adding source information and this IVB mesh object either use `LayerMesh` directly or derive from it if additional functionality is needed

parastell/parastell.py
connoramoreno: We can, I just limited this method to one component per mesh for simplicity. Depending on how we would want to implement multi-component meshes, it could get rather complicated. We could just extend the current algorithm to loop over the requested components, but that would result in overlapping tets and repeated vertices for adjacent components.

To avoid overlapping tets, we could alternate hex splitting routines like we do for the source mesh, but that would require that users input an ordered list of components, and that we check that order (for guaranteed correct results, anyways). To avoid repeated vertices, we could include logic to ensure that, in the case of adjacent components, common vertices are shared. I'm just worried the logic could get a little convoluted.

It just feels like it could be a bit of an unnecessarily heavy lift considering that single-tet thick meshes are likely only going to be used for single-component mesh tallies anyways. If there's interest in it, though, I can work on that modification.

parastell/invessel_build.py
connoramoreno: I also remember that there's a reason we went this route in the past, but I also don't remember why.

parastell/invessel_build.py
connoramoreno: Perhaps it would be more appropriate to change these variable names, but I defined these separately because it can be the case that `self.num_ribs` and `self._num_ribs`, etc. don't have the same value, due to the way we expand the angle lists. For the MOAB routine, it's important we know the actual number of ribs and rib points, rather than what the user defined.

parastell/invessel_build.py
connoramoreno: I'm sure we can find a way to make this method compatible with both method mesh routines.

parastell/parastell.py
gonuke: I think it's fine to leave it this way - maybe just add a comment in the docstring about the risks of calling this on multiple components.... as you outlined in your response above

gonuke (approved):
LGTM, but needs a rebase....

connoramoreno:
Oh, I wanted to make some changes to the MOAB workflow before merging (implementing a common derived class for `InVesselBuild` and `SourceMesh`)

gonuke:
> Oh, I wanted to make some changes to the MOAB workflow before merging (implementing a common derived class for `InVesselBuild` and `SourceMesh`)

Do you want to convert this to draft, then?

connoramoreno:
Not sure why the Docker deployment workflow was triggered? This PR doesn't change dependencies at all...

parastell/source_mesh.py
```diff
@@ -282,18 +281,20 @@ def create_vertices(self):
 
                     vert_idx += 1
 
-        self.verts = self.mbc.create_vertices(self.coords)
+        self.add_vertices(self.coords)
 
-    def _source_strength(self, tet_ids):
+    def _compute_tet_data(self, tet_ids, tet):
         """Computes neutron source strength for a tetrahedron using five-node
```
gonuke, line 287: Add something to the docstring to clarify that this method sets the value of the tag, as well...

gonuke (changes requested):
This looks pretty good now.  Just two little comments and should be ready to merge

parastell/source_mesh.py
```diff
@@ -375,166 +361,56 @@ def _get_vertex_id(self, vertex_idx):
...
-                    self._create_tets_from_hex(
+                    tets, vertex_id_list = self._create_tets_from_hex(
                         cfs_idx, poloidal_idx, toroidal_idx
                     )
+                    [
+                        self._compute_tet_data(tet_ids, tet)
+                        for tet_ids, tet in zip(vertex_id_list, tets)
+                    ]
 
     def export_mesh(self, filename="source_mesh", export_dir=""):
```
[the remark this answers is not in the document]
connoramoreno: True, but I'd like to keep that logger message and I'm unsure how it could be incorporated well into the parent class method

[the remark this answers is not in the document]
connoramoreno: I suppose I could just add a logger object to `ToroidalMesh` and make the export message more generic

gonuke (approved):
Thanks @connoramoreno

PR #212, opened 2025-04-07
update to import pydagmc

### LINKED ISSUE #211: Prepare for renaming of dagmc to pydagmc
https://github.com/svalinn/pydagmc/pull/31

`import dagmc` is becoming `import pydagmc`, it will be necessary to make a few syntax updates.

Edgar-21:
Updates to use the new import name for pydagmc.
Closes #211
Tests should be failing currently, when https://github.com/svalinn/pydagmc/pull/31 gets merged, we should be able to update the CI image and see that tests will be passing.

### DIFF
--- a/parastell/invessel_build.py
+++ b/parastell/invessel_build.py
@@ -6,7 +6,7 @@ from scipy.interpolate import RegularGridInterpolator
 
 import cadquery as cq
 import pystell.read_vmec as read_vmec
-import dagmc
+import pydagmc
 from pymoab import core, types
 
 from . import log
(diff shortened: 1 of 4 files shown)


Edgar-21:
I just tried triggering the CI image to rebuild with the hope that it would grab the new version of PyDAGMC but it seems like it must have used a cached version because the tests all still passed on main.

gonuke (approved):
Thanks @Edgar-21

gonuke:
I manually launched the `docker_publish` workflow on this branch and it passed, so I'll go ahead.

PR #226, opened 2025-05-07
Import all solids from pre-generated magnet step files

Edgar-21:
As previously written, only a single solid was being meshed and added to the DAGMC model. The test magnet only has a single volume, so we snuck through testing.


gonuke (approved):

gonuke:
Should we update tests?

connoramoreno:
I don't believe tests are affected by this change/issue

gonuke:
> I don't believe tests are affected by this change/issue 

I got that, but you suggest that this bug existed because of inadequate testing...  maybe we add a test that would have failed before this?

MicahGale:
I agree with @gonuke. My personal rule is that any time I find a bug your first commit should be making/modifying a test that fails due to the bug. That way I know I _actually_ fixed the bug. (Or at least 1 variant of the bug).

gonuke:
Doh! it looks like the changes to this STEP file broke a bunch of other tests... simplest may be to make 2 step files, one with only a single magnet and one with 2 magets - rather than change all the other tests to get this right.

Edgar-21:
not a big deal to change the other tests, just a few constants. I was delayed by finding out the newest version of cad-to-dagmc breaks stuff as well, which was the source of many of the failures in previous run. I've pinned the version of cad-to-dagmc to one that works as a temporary measure, rather than expanding this PR to fix that as well. There also seems to be an issue with pyqt5 and pyqt5-sip in our current testing environment that  is causing issues with the NWL tests.

Edgar-21:
actually looks like the NWL thing was only a problem on my local install. Weird.

environment.yml
```diff
@@ -8,7 +8,7 @@ dependencies:
   - scipy
   - cadquery
   - moab=5.5.1[build=nompi_tempest_*]
-  - cad_to_dagmc >=0.8.2
+  - cad_to_dagmc=0.8.2
```
connoramoreno, line 11:
```suggestion
  - cad_to_dagmc >=0.8.2, <=0.9.0
```

connoramoreno (changes requested):
One small suggestion about CAD-to-DAGMC version pinning to give more flexibility to the environment, and I'll be happy to merge (thank you for contributing during your fun-employment)

connoramoreno (approved):
Actually, I'll be submitting a new PR soon to update our API for the new version of CAD-to-DAGMC, pinning 0.8.2 is fine for the time being. Thanks @Edgar-21!

PR #228, opened 2025-05-14
Update syntax for newer versions of CAD-to-DAGMC

### LINKED ISSUE #227: Cad-to-dagmc API changes
It seems that with the release of [0.9.1](https://github.com/fusion-energy/cad_to_dagmc/releases/tag/0.9.1) there was a change to the API breaking the current implementation in Parastell. It seems like there might be opportunity to simplify the cad-to-dagmc workflow using the new features in this release.

I've only done a surface level investigation but the main thing is that `mesh_brep()` is no longer a method as of 0.9.1. Not sure if it was renamed or split up or what.

connoramoreno:
Updates use of CAD-to-DAGMC's API for versions 0.9.1+ and pins those versions in the conda environment file.

Closes #227

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -8,7 +8,7 @@ dependencies:
   - scipy
   - cadquery
   - moab=5.5.1[build=nompi_tempest_*]
-  - cad_to_dagmc=0.8.2
+  - cad_to_dagmc >=0.9.1
   - ca-certificates
   - certifi
   - openssl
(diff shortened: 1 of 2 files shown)


MicahGale:
This alone will only affect people doing a clean install. Migration instructions for people with it currently installed for how to update their conda environment should probably be included in the release notes I think.

gonuke (approved):
LGTM - we should definitely think about how we guide people to update their environment, in general, but that could be a different PR

PR #229, opened 2025-05-14
Add support for continuous, 360-degree geometries in PyDAGMC workflow

connoramoreno:
Adds support for modeling continuous, 360-degree geometries for the PyDAGMC workflow (note that continuous geometries were already supported in the CAD-based workflow using the `repeat` parameter). Also adds a test to `test_parastell.py` to check this functionality.

### DIFF
--- a/parastell/invessel_build.py
+++ b/parastell/invessel_build.py
@@ -490,6 +490,7 @@ class InVesselBuild(object):
 
     def _connect_ribs_with_tris_moab(self, rib1, rib2, reverse=False):
         """Creat MBTRI elements add add them to a surface between two ribs.
+        (Internal function not intended to be called externally)
 
         Arguments:
             rib1 (Rib object): First of two ribs to be connected.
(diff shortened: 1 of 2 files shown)


parastell/invessel_build.py
```diff
@@ -514,25 +515,38 @@ def _connect_ribs_with_tris_moab(self, rib1, rib2, reverse=False):
...
         """
         self.curved_surface_ids = []
         surfaces = list(self.Surfaces.values())
         first_surface = surfaces[0]
         for surface in surfaces:
             mb_tris = []
-            for rib, next_rib in zip(surface.Ribs[0:-1], surface.Ribs[1:]):
+
+            if continuous_geom:
+                ribs = surface.Ribs[:-1] + [surface.Ribs[0]]
```
gonuke, line 545: This throws out the last rib - do we assume that it should be the same as rib 0 for a continuous geometry? Or do we assume that we don't care what it is?

parastell/invessel_build.py
```diff
@@ -622,24 +646,24 @@ def _tag_volumes_with_materials_pydagmc(self):
...
-            self._logger.error(e.args[0])
-            raise e
         self._logger.info(
             "Generating DAGMC model of in-vessel components with PyDAGMC..."
         )
+
+        if (
+            self.radial_build.toroidal_angles[-1]
+            - self.radial_build.toroidal_angles[0]
+            == 360.0
```
gonuke, line 656: Do we want to test for this to be some version of close to 360 rather than exactly 360? Like we originally did...

parastell/invessel_build.py
```diff
@@ -622,24 +646,24 @@ def _tag_volumes_with_materials_pydagmc(self):
...
+        ):
+            continuous_geom = True
```
gonuke, line 658: For some reason `continuous_geom` isn't an obvious name to me.  My first thought was `closed_geom` but I guess a single period with end caps is also closed... `full_360` or something like that??? I'm not feeling very creative...

gonuke (changes requested):
A couple of minor things before I check the tests...

parastell/invessel_build.py
connoramoreno: We assume it should be the same as rib 0

parastell/invessel_build.py
connoramoreno: Yes, that's my bad

gonuke (approved):
LGTM - thanks @connoramoreno

PR #230, opened 2025-05-14
Add support for custom source mesh grids

connoramoreno:
Adds support for users to input custom flux-coordinate grids to define source mesh vertices. This replaces the previous system which only enabled uniformly spaced (in flux-coordinates) grid points along each coordinate axis.

### DIFF
--- a/parastell/parastell.py
+++ b/parastell/parastell.py
@@ -389,19 +389,20 @@ class Stellarator(object):
             filename=filename, export_dir=export_dir
         )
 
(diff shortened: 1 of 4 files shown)


gonuke (changes requested):
This looks pretty good to me @connoramoreno.  A few minor comments here.

Perhaps we should add a test with a non-uniform set of grid points in one or more dimensions?

gonuke (approved):
Thanks @connoramoreno - LGTM

PR #231, opened 2025-05-15
Add support for multi-component IVB meshes using the MOAB workflow

connoramoreno:
Enables the generation of meshes including multiple components for the MOAB workflow. This workflow avoids repeated vertices for adjacent components while supporting gaps from excluded components.

### DIFF
--- a/parastell/invessel_build.py
+++ b/parastell/invessel_build.py
@@ -772,10 +772,11 @@ class InVesselBuild(object):
         for component_idx, component in enumerate(components):
             # Extract inner and outer surfaces of current component
             outer_surface = self.Surfaces[component]
-            # Inner surface of current component is outer surface of the previous
-            # component. Since surfaces are created in order of components and
-            # named after the component for which they are the outer surface,
-            # it can be found by the ordered list of surface keys
+            # Inner surface of current component is outer surface of the
+            # previous component. Since surfaces are created in order of
+            # components and named after the component for which they are the
+            # outer surface, it can be found by the ordered list of surface
+            # keys
             inner_surf_idx = surface_keys.index(component) - 1
             inner_component = surface_keys[inner_surf_idx]
             inner_surface = self.Surfaces[inner_component]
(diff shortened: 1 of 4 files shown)


parastell/invessel_build.py
```diff
@@ -713,27 +713,61 @@ def extract_solids_and_mat_tags(self):
...
-        component_idx = surface_keys.index(component)
-        surfaces = [self.Surfaces[surface_keys[component_idx - 1]]]
-        surfaces.append(self.Surfaces[component])
+        # Identify surfaces and gaps in mesh
+        for component_idx, component in enumerate(components):
+            component_surf_idx = surface_keys.index(component)
+            previous_component = surface_keys[component_surf_idx - 1]
+
+            # Conditionally handle gap in mesh
+            if previous_component != components[component_idx - 1]:
```
gonuke, line 761: What do you expect when `component_idx = 0`?

gonuke: Maybe handle the first component in a special way and loop from `components[1:]`

gonuke, line 761: This assumes that `components` is sorted in the same way that the components were built.  Either error or warning and/or sort them appropriately.

parastell/invessel_build.py
```diff
@@ -1123,13 +1177,14 @@ def create_vertices(self):
...
     def create_mesh(self):
         """Creates volumetric mesh in real space."""
-        surface_idx = 0
-
-        for toroidal_idx in range(self._num_ribs - 1):
-            for poloidal_idx in range(self._num_rib_pts - 1):
-                self._create_tets_from_hex(
-                    surface_idx, poloidal_idx, toroidal_idx
-                )
+        for surface_idx, _ in enumerate(self._surfaces[:-1]):
```
gonuke, line 1180:
```suggestion
        for surface_idx, _ in enumerate(self.surfaces[:-1]):
```

parastell/invessel_build.py
```diff
@@ -713,27 +713,61 @@ def extract_solids_and_mat_tags(self):
...
+            self._logger.warning(e.args[0])
+            components.remove(component)
+
+        if "plasma" in components:
+            remove_inner_component("plasma")
+        elif "chamber" in components:
+            remove_inner_component("chamber")
+
         surfaces = []
+        gap_map = []
```
gonuke, line 751: Document what the list of `surfaces` will be and how it relates to `gap_map`.

PR #235, opened 2025-06-16
Ensure volumetric meshes match DAGMC meshes in corresponding routines

### LINKED ISSUE #234: Volumetric meshes not conformal to DAGMC surface meshes
There is currently no way to ensure that the surface boundaries of IVB and magnet volumetric meshes will match/be conformal to DAGMC surface meshes. It would be good to ensure this for each meshing routine and its "corresponding" DAGMC routine:

- CAD-based Gmsh and CAD-to-DAGMC
- Cubit
- PyDAGMC-based Gmsh and PyDAGMC
- PyMOAB and PyDAGMC

connoramoreno:
This branch modifies the IVB and magnet volumetric meshing workflows such that their surface boundaries match corresponding DAGMC surface meshes when the same meshing parameters are given.

- CAD-based Gmsh IVB and magnet volumetric meshes now match CAD-to-DAGMC surface meshes
- Cubit IVB and magnet volumetric meshes now match Cubit DAGMC surface meshes, with caveats:
  - Some minor exceptions have been observed for triangles on the end caps of in-vessel components, but the effect is hopefully unimportant since the mesh geometry is still conformal. 
  - Magnet volumetric meshing can be finicky if the meshing parameters are not well-set
- PyDAGMC-based Gmsh IVB volumetric meshes now match PyDAGMC surface meshes

A notable exception here is that PyMOAB IVB volumetric meshes **do not** match PyDAGMC surface meshes. For PyDAGMC geometries, it was chosen to have triangles oriented the same way for adjacent surfaces, to prevent overlaps in the DAGMC geometries. However, the hex-splitting routine that the PyMOAB routine uses ensures that triangles on adjacent surfaces are oriented oppositely. To remedy this, either:

- The PyDAGMC routine would need to be changed, such that the triangle creation routine matches the surface boundary in the PyMOAB routine. This risks triangle overlap in meshes where components are thin relative to the resolution of the mesh (note that triangles/tetrahedra overlap in this case already for the PyMOAB routine)
- The PyMOAB routine would need to be changed, such that the hex-splitting routine matches the triangle creation routine in the PyDAGMC routine. This would likely introduce meshes with more elements, since an alternative hex-splitting routine would likely introduce more tetrahedra per hexahedron.

We can also leave both routines as-is since the PyDAGMC-Gmsh workflow supports conformal meshes.

Closes #234

### DIFF
--- a/parastell/magnet_coils.py
+++ b/parastell/magnet_coils.py
@@ -12,7 +12,6 @@ from .cubit_utils import (
     create_new_cubit_instance,
     import_geom_to_cubit,
     export_mesh_cubit,
-    mesh_volume_skeleton,
     mesh_volume_auto_factor,
     mesh_surface_coarse_trimesh,
     get_last_id,


parastell/utils.py
```diff
@@ -255,71 +255,45 @@ def smooth_matrix(matrix, steps, sigma):
...
+                for Quads, 9: Packing of Parallelograms, 11: Quasi-structured
+                Quad.
         filename (str): path to input mesh file. Must be a Gmsh-compatible file
             type. Options include VTK and MSH file types.
 
     Returns:
         filename (str): path to remeshed mesh output file. The output file type
             and path will be the same as the input.
     """
-    # These parameters are taken directly from Gmsh documentation (see
```
gonuke, line 274: Curious: was all of this deleted code doing nothing important?

parastell/magnet_coils.py
```diff
@@ -13,6 +13,8 @@
     import_geom_to_cubit,
     export_mesh_cubit,
     mesh_volume_skeleton,
```
gonuke, line 15: Can we delete this? And it seems like it won't be used anymore, should we delete the method from `cubit_utils`?

parastell/cubit_utils.py
```diff
@@ -278,15 +239,19 @@ def merge_surfaces(surface_1, surface_2):
...
 
 
-def mesh_volume_auto_factor(volume_ids, mesh_size=5.0):
+def mesh_volume_auto_factor(volume_ids=None, mesh_size=5.0):
     """Meshes a volume in Cubit using automatically calculated interval sizes.
 
     Arguments:
-        volume_ids (iterable of int): Cubit IDs of volumes to be meshed.
+        volume_ids (iterable of int): Cubit IDs of volumes to be meshed
+            (defaults to None).
```
gonuke, line 247:
```suggestion
            (defaults to None) if no IDs are provided, all are meshed.
```

gonuke (changes requested):
This mostly looks good - just a couple of questions, including my own education

parastell/magnet_coils.py
connoramoreno: Although, I suppose I could remove it now and if I want to restore it, I can just copy-paste from an old commit

gonuke: I'm fine to leave it there for posterity... is it "cleaned up" with respect to the standard meshing parameters that you've used elsewhere?

PR #236, opened 2025-07-02
Enable use of rib-based surfaces in NWL workflow

connoramoreno:
This branch updates the NWL utility to enable the use of rib-based surface geometries. With these changes, the NWL workflow uses either the `VMECSurface` or the `RibBasedSurface` class as the basis for its root-solving.


parastell/nwl_utils.py
```diff
@@ -145,19 +149,20 @@ def solve_poloidal_angles(data):
         result = direct(
             compute_residual,
             bounds=[(0, 2 * np.pi)],
-            args=(vmec_obj, wall_s, toroidal_angle, point),
+            args=(ref_surf, wall_s, toroidal_angle, point),
```
gonuke, line 152: I would assume that `wall_s` is no longer relevant for a surface that is not defined by VMEC.  Do we assume that a user sets this to 0? or should we set the value of `wall_s` inside the conditional above with different values for each case?

gonuke (changes requested):
This looks like it was pretty straightforward - thanks @connoramoreno .  I do have one question about how we treat `wall_s` in the case where we use a rib-based surface.  I'm not sure how we use that parameter when we make an IVB, but we should be consistent here.

parastell/nwl_utils.py
connoramoreno: That's correct, `wall_s` is not relevant for anything other than a VMEC surface. Its value does not matter, and in fact is not even used, if the user is not using a VMEC surface. It's just a placeholder to make `angles_to_xyz` input the same between all surface classes defined in `invessel_build.py`.

I toyed with changing inputs such that `wall_s` was only used for `VMECSurface` but ended up figuring that was best left to a separate PR.

gonuke (approved):
Thanks @connoramoreno

PR #238, opened 2025-07-03
Actually add pyproject

MicahGale:
Migration from #200.


pyproject.toml
```diff
@@ -0,0 +1,77 @@
...
+	"numpy~=1.26.4",
+	"scipy",
+	"cadquery",
+	"pymoab~=5.5.1",
+	"cad_to_dagmc>=0.9.1",
+	"openmc>=0.15.0",
+	"matplotlib",
+	"netcdf4",
+	"pyyaml",
+	"pystell",
```
connoramoreno, line 51: I think we'll want to vendor PyStell before we merge this

connoramoreno (commented):
This looks good to me, we'll just need to hold off on merging this until we vendor PyStell

MicahGale:
With `pystell` now vendored I believe we are now just waiting on the PyDagmc deployment to PyPI.

gonuke:
I now have a PyPI account and have requested both Svalinn and CNERG orgs... if/when those orgs are approved, I can figure out next steps (including possibly adding others to those orgs and letting them do the work 🤣 )

MicahGale:
Also if the org request is denied I discovered that you can have multiple owners (highest permission) on PyPI, so this can be transferred around.

MicahGale:
I'm running into issues with the wrong python versions being installed. I can't remember: do we still need to pin python to 3.11 @connoramoreno? Also this should be specified in the `environment.yaml` as well.

connoramoreno:
@MicahGale I removed the Python pin when we vendored PyStell

MicahGale:
Was pystell the one holding us back to 3.11 only? Can I safely say `>=3.11`?

connoramoreno:
Correct and yes you can, but why `>= 3.11`?

pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+	"numpy~=1.26.4",
+	"scipy",
+	"cadquery",
+	"pymoab~=5.5.1",
+	"cad_to_dagmc>=0.9.1",
+	"openmc>=0.15.0",
+	"matplotlib",
+	"netcdf4",
+	"pyyaml",
+      	"pydagmc"
```
gonuke, line 51: Indent inconsistency 
```suggestion
  "pydagmc"
```

Dockerfile
```diff
@@ -44,6 +44,7 @@ RUN cp /root/.bashrc /opt/etc/bashrc
 # Install Python dependencies in parastell env conda environment
 COPY ./environment.yml /environment.yml
 RUN conda env create -f environment.yml
+RUN pip install .[develop]
```
gonuke, line 47: What does the `[develop]` mean?

gonuke (commented):
I think these are all for my education/clarification...

environment.yml
```diff
@@ -3,15 +3,11 @@ channels:
   - conda-forge
 dependencies:
   - cadquery
-  - moab >=5.5.1[build=nompi_tempest_*]
-  - cad_to_dagmc >=0.9.1
-  - openmc >=0.15.0[build=dagmc_nompi_*]
-  - numpy
-  - scipy
```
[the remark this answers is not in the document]
MicahGale: No; they got moved to `pyproject.toml` and so can be pip installed with `parastell`. I'm realizing you are probably against this approach for mixing pip and conda.

pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+]
+
+[project.optional-dependencies]
+test = ["pytest", "pytest-cov"]
+
+format = ["black==25.*"]
+
+build = ["build"]
+
+develop = ["parastell[test,format,build]"]
```
MicahGale, line 61: @gonuke  this is where the optional dependencies are specified for `.[develop]` or `parastell[develop]`. The optional specifications can be a list, and this dependency is recursive to by DRY.

Dockerfile
MicahGale: This specifies installing an optional dependency option, so my other comment.

environment.yml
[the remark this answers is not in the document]
connoramoreno: If the intent is to move to conda in the long-term, maybe it's best to keep them (plus, it seems harmless to do so)? I'm unsure what the conda infrastructure will have to look like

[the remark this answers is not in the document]
MicahGale: For the conda infrastructure the idea is that all of this would be moved to separate conda forge repo.

pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+       "Natural Language :: English",
+       "Operating System :: OS Independent",
+       "Topic :: Scientific/Engineering :: Physics",
+       "Topic :: Scientific/Engineering",
+       "Topic :: Scientific/Engineering :: Human Machine Interfaces",
+       "Programming Language :: Python :: 3 :: Only",
+       "Programming Language :: Python :: 3.11",
+]
+
+dependencies = [
```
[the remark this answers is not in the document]
MicahGale: It doesn't. It just sees if a package is installed; if not, it will install it itself. This is why the conda environment needs to be setup first.

gonuke (approved):
I haven't tested this locally, but since the new build CI test is passing, it seems like this can be approved.  Maybe @connoramoreno can test on a clean docker image, or similar, to confirm that it works for users per the given instructions?

pyproject.toml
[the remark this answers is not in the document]
connoramoreno: Oh right

environment.yml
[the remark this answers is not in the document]
gonuke: I'm ok with the mix, it's proven necessary for the time being. It's unclear which way the community of dependencies will shift over time anyway

connoramoreno:
Can confirm the installation process works on a clean Docker image (a version of ParaStell's Docker image that doesn't include Cubit), although I'm getting a segmentation fault during testing. Not sure what's going on there but it's probably something to do with my specific Docker image given that all tests pass here. I can run a different ParaStell script just fine. I just have some minor changes to request, mostly related to documentation. I'll submit another review soon

environment.yml
```diff
@@ -3,15 +3,11 @@ channels:
...
+  - moab>=5.5.1[build=nompi_tempest_*]
+  - ca-certificates
+  - certifi
+  - openssl
+  - openmc>=0.15.0[build=dagmc_nompi_*]
+  - cad_to_dagmc>=0.9.1
   - matplotlib
```
connoramoreno, line 12: Nitpicky but I prefer the inclusion of `numpy` and `scipy`, and the omission of `ca-certificates`, `certifi`, and `openssl`. I suppose if we wanted to be minimalistic, though, we would preserve the omission of `numpy` and `scipy`, and also omit `matplotlib` below (dependency of OpenMC).
```suggestion
  - moab>=5.5.1[build=nompi_tempest_*]
  - cad_to_dagmc>=0.9.1
  - openmc>=0.15.0[build=dagmc_nompi_*]
  - numpy
  - scipy
  - matplotlib
```
or
```suggestion
  - moab>=5.5.1[build=nompi_tempest_*]
  - cad_to_dagmc>=0.9.1
  - openmc>=0.15.0[build=dagmc_nompi_*]
```

README.md
```diff
@@ -61,24 +61,18 @@ While it is possible to use ParaStell with older versions of Cubit, additional s
...
+### Finally Install Parastell
 
-To make use of this feature, add the ParaStell `executables` directory to your `PATH` by adding a line similar to the following to your `.bashrc` file:
+Now that all dependencies have been installed, you can install ParaStell with `pip`. Run the following command from the root of the ParaStell repository:
 
-```bash
-export PATH = $PATH:$HOME/parastell/executables/
+``` bash
+pip install .
 ```
```
connoramoreno, line 70:
```suggestion
``` bash
pip install .
```

Or, to install in develop mode for `pytest` support,
```bash
pip install .[develop]
```

pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+	"numpy~=1.26.4",
+	"scipy",
+	"cadquery",
+	"pymoab~=5.5.1",
+	"cad_to_dagmc>=0.9.1",
+	"openmc>=0.15.0",
+	"matplotlib",
+	"netcdf4",
+	"pyyaml",
+        "pydagmc"
```
connoramoreno, line 51: I image the conda dependencies included in this list should match those in `environment.yml` (packages and their versions)

connoramoreno (changes requested):
Some final comments, thanks @MicahGale

connoramoreno:
Oh, I forgot to include in my review: @MicahGale I should've done this in a previous PR but can you update `INSTALL.md` to reflect our changes in dependencies and (Python, Python packages, and versions)?

README.md
gonuke: I think github has gotten lost on the nested "```" so you may want to incorporate this suggestion carefully.

environment.yml
MicahGale: I want to do more research before messing with `ca-certificates` and `openssl`. One does not simply change these sort of things.

README.md
connoramoreno: Yeah, I thought I counted an incorrect number of "```"s but this was the only way to get GitHub to render the preview properly. Probably better to make this change manually.

MicahGale: Yes, I'm not following, could you remove the wrapping `suggestions`, and just provide what you think it should say?

connoramoreno:
``` bash
pip install .
```

Or, to install in develop mode for `pytest` support,
```bash
pip install .[develop]
```

pyproject.toml
MicahGale: Yes it should be as permissive as possible idealy.

gonuke:
> > Are there users who don't want to the `environment.yml` file?
> 
> 
> 
> Not sure! Just covering bases. I suppose if people really want to install "manually" (i.e., not via `environment.yml`), they probably don't need a special guide and would know to refer to `environment.yml` for conda dependencies. Thoughts @gonuke?

Upon a quick skim, it seems a little redundant with what's now in README.  I'd say that it either needs to be more extensive or removed.

connoramoreno:
I'm fine just to remove it then

connoramoreno (changes requested):
Last few comments; just minor changes and some questions! Thank you @MicahGale

pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+       "Operating System :: OS Independent",
+       "Topic :: Scientific/Engineering :: Physics",
+       "Topic :: Scientific/Engineering",
+       "Topic :: Scientific/Engineering :: Human Machine Interfaces",
+       "Programming Language :: Python :: 3 :: Only",
+       "Programming Language :: Python :: 3.11",
+]
+
+dependencies = [
+    "numpy>=1.26.4",
```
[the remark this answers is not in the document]
connoramoreno: We specify version requirements, e.g., `cad_to_dagmc` and `moab`, in `environment.yml`. Are they also required here?

PR #241, opened 2025-07-08
Relax dependencies

connoramoreno:
Relaxes the dependencies specified in `environment.yml`. One major change is that we no longer require a specific version of Python


gonuke (approved):
It looks like it still works...

PR #242, opened 2025-07-22
Add tetrahedral volume attribute to `InVesselComponentMesh` class

connoramoreno:
This branch adds an attribute to the `InVesselComponentMesh` class that stores tetrahedral volumes in an ordered list, as is done for the `SourceMesh` class. The tetrahedral volume calculation, originally done in the `_compute_tet_data` method of the `SourceMesh` class, was isolated to a new `_compute_tet_volume` method and moved to the `ToroidalMesh` abstract base class.

### DIFF
--- a/parastell/invessel_build.py
+++ b/parastell/invessel_build.py
@@ -1237,6 +1237,10 @@ class InVesselComponentMesh(ToroidalMesh):
         self.surfaces = surfaces
         self.gap_map = gap_map
 
+        self.volumes = []
+
+        self._add_tags_to_core()
+
     @property
     def surfaces(self):
         return self._surfaces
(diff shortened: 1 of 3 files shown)


parastell/invessel_build.py
```diff
@@ -1265,14 +1269,45 @@ def gap_map(self, list):
...
         for surface in self.surfaces:
             for rib in surface.Ribs:
-                coords.extend(rib.rib_loci)
-        coords = np.array(coords)
-        self.add_vertices(coords)
+                self.coords.extend(rib.rib_loci)
+        self.coords = np.array(self.coords)
+        self.add_vertices(self.coords)
+
+    def _compute_tet_data(self, tet_ids, tet):
```
gonuke, line 1298: since tets also have ids, I think it would be better to call this `vert_ids` (here and everywhere else).  Also maybe a more clear function name (although long!)
```suggestion
    def _compute_and_tag_tet_volume(self, vert_ids, tet):
```

parastell/invessel_build.py
```diff
@@ -1265,14 +1269,45 @@ def gap_map(self, list):
...
+        """Computes tetrahedron volume, and sets the corresponding value of
+        the respective tag for that tetrahedron.
+        (Internal function not intended to be called externally)
+
+        Arguments:
+            tet_ids (list of int): tetrahedron vertex indices.
+            tet (object): pymoab.EntityHandle of tetrahedron.
+        """
+        tet_vol = self._compute_tet_volume(tet_ids)
+        self.volumes.append(tet_vol)
```
gonuke, line 1308: Seems a little buried to update this list here.  Could you return the volume and then append it elsewhere?

parastell/invessel_build.py
```diff
@@ -1281,9 +1316,13 @@ def create_mesh(self):
...
             for toroidal_idx in range(self._num_ribs - 1):
                 for poloidal_idx in range(self._num_rib_pts - 1):
-                    self._create_tets_from_hex(
+                    tets, vertex_id_list = self._create_tets_from_hex(
                         surface_idx, poloidal_idx, toroidal_idx
                     )
+                    [
+                        self._compute_tet_data(tet_ids, tet)
+                        for tet_ids, tet in zip(vertex_id_list, tets)
+                    ]
```
gonuke, line 1325: If `_compute_tet_data` returned the `tet_vol` you could `self.volumes.extend()` by the list you create here.

gonuke (changes requested):
a little naming and separation of concerns

PR #244, opened 2025-07-23
Ensure Kisslinger data is formatted as expected

connoramoreno:
Adds a routine to format Kisslinger data as ParaStell expects. This routine is called automatically in the `ribs_from_kisslinger_format` function. Also updates tests.


parastell/invessel_build.py
```diff
@@ -133,15 +133,26 @@ def angles_to_xyz(self, toroidal_angle, poloidal_angles, s, scale):
...
         rib_data (numpy array): NxMx2 array of of R, Z points. The first
             dimension corresponds to the plane of constant toroidal angle on
             which the closed loop of points lies. The second dimension is the
             location on the closed loop at which the point lies, and the third
-            dimension is the R, Z values of that point.
+            dimension is the R, Z values of that point. ParaStell expects the
+            following from this data set:
+            - The data spans exactly one field period
+            - The coordinates of each toroidal slice (rib) precess counter-
+              clockwise
```
gonuke, line 148: I'm not sure that "counter-clockwise" is well defined here.  It depends whether you are looking "forward" or "backward" along the magnetic axis...  I think we want them to precess upwards at the outboard and downwards at the inboard?

tests/test_utils.py
```diff
@@ -124,34 +133,117 @@ def test_ribs_from_kisslinger_format():
...
+        _,
+        _,
+        _,
+        _,
+        original_surface_coords,
+    ) = ribs_from_kisslinger_format(
+        original_ribs_file,
+        delimiter=" ",
+        scale=1,
+    )
```
gonuke, line 155: Since this is only used in a test near the end of this method, perhaps move this function call there.  Might also name the returned data as `expected_surface_coords`?

tests/test_utils.py
```diff
@@ -124,34 +133,117 @@ def test_ribs_from_kisslinger_format():
...
+        unscrambled_surface_coords[0] - unscrambled_surface_coords[-1], 0
     )
     assert not np.allclose(
-        custom_surface_rz_ribs[0], custom_surface_rz_ribs[1]
+        unscrambled_surface_coords[0], unscrambled_surface_coords[1]
     )
+    assert np.allclose(unscrambled_surface_coords, original_surface_coords)
+
+
+def test_stl_surfaces_to_cq_solid():
```
gonuke, line 187: The diffs always get messed up when you move functions around like this :)

parastell/magnet_coils.py
```diff
@@ -521,32 +522,21 @@ def coords(self, data):
...
-        # have opposite signs)
-        shifted_coords = np.append(coords[1:], [coords[1]], axis=0)
-        midplane_flags = -np.sign(coords[:, 2] * shifted_coords[:, 2])
-        # Find index of outboard midplane point
-        outboard_index = np.argmax(midplane_flags * radii)
+        radii = np.linalg.norm(self.coords[:, :2], axis=1)
+        z_values = self.coords[:, 2]
+
+        # Reformulate coordinates as R, Z pairs
+        coords = np.array(list(zip(radii, z_values)))
```
gonuke, line 537: I'm assuming you can't go directly from `zip` to `np.array()`? hmm....

parastell/utils.py
```diff
@@ -227,32 +285,101 @@ def reorder_loop(list, index):
...
+            closed loop.
+        positive (bool): progress coordinates in positive z-direciton
+            (defaults to True). If negative, coordinates will progress in
+            negative direction.
+
+    Returns:
+        coords (Nx2 numpy.array of float): reordered list of R, Z coordinates
+            for the closed loop.
+    """
+    if positive == (coords[0, -1] > coords[1, -1]):
```
gonuke, line 379: Does this assume that the first entry is on the outboard?  I am not sure I see all the places this is used, but if the first entry is on the inboard, then we want `positive = False`.  Not sure how we test that?

gonuke (changes requested):
A few comments here... did you intentionally rearrange the order of the functions? Or does black do that?  Every time you insert functions and move other functions, it bloats the PR and makes it hard to review.

connoramoreno:
Sorry for all the bloat - didn't realize it would make review difficult

parastell/utils.py
```diff
@@ -46,14 +95,30 @@ def enforce_helical_symmetry(matrix):
...
     """
-    num_rows, num_columns = matrix.shape
+    original_shape = matrix.shape
+
+    if len(original_shape) == 2:
+        num_rows, num_columns = original_shape
+        flattened_shape = num_rows * num_columns
+    elif len(original_shape) == 3:
+        num_rows, num_columns, num_coords = original_shape
+        flattened_shape = (num_rows * num_columns, num_coords)
```
connoramoreno, line 115: Define `coordinate_data` (or similar) boolean for future tests

parastell/utils.py
```diff
@@ -64,24 +129,66 @@ def enforce_helical_symmetry(matrix):
...
+        last_half[:, -1] *= -1
+
     flattened_matrix = np.concatenate([first_half, last_half])
 
-    matrix = flattened_matrix.reshape((num_rows, num_columns))
+    matrix = flattened_matrix.reshape(original_shape)
+
+    if len(original_shape) == 3:
+        # Ensure periodicity at ends of period
+        matrix[-1] = matrix[0]
```
connoramoreno, line 159: Expand documentation for coordinate array case, remove conditional

tests/test_utils.py
```diff
@@ -82,6 +48,143 @@ def test_expand_list():
...
+    expected shape, and contain the expected values, by testing if:
+      * the returned matrix has the same 2-D or 3-D shape as the input
+      * the returned matrix has the expected values
+    """
+    matrix_symmetric = enforce_helical_symmetry(matrix)
+    assert matrix_symmetric.shape == shape_exp
+    assert np.allclose(matrix_symmetric, matrix_exp)
+
+
+def test_ribs_from_kisslinger_format():
```
connoramoreno, line 124: Also test formatting of original ribs file?

parastell/utils.py
```diff
@@ -46,14 +95,33 @@ def enforce_helical_symmetry(matrix):
         - Matrix represents a full stellarator period
         - The first row corresponds to a toroidal angle at the beginning of a
           period (i.e., poloidal symmetry is expected)
+        - The beginning, toroidal midplane, and end of the period are symmetric
+          about the axial midplane
+        - If a 3-D matrix, the final axis represents R, Z coordinates.
```
gonuke, line 100:
```suggestion
        - The matrix values can be scalars, thus appearing 
          as a 2-D matrix.  Alternatively, those values 
          can be vectors of coordinates, thus appearing 
          as a 3-D matrix.
```

parastell/utils.py
```diff
@@ -46,14 +95,33 @@ def enforce_helical_symmetry(matrix):
...
+        flattened_shape = num_rows * num_columns
+        coordinate_data = False
+    elif len(original_shape) == 3:
+        num_rows, num_columns, num_coords = original_shape
+        flattened_shape = (num_rows * num_columns, num_coords)
+        coordinate_data = True
+    else:
+        e = ValueError(
+            f"Input matrix is of shape {original_shape}, but only 2-D and 3-D "
+            "arrays are supported."
```
gonuke, line 122:
```suggestion
            f"Input matrix is of shape {original_shape}, but only 2-D arrays of either "
            "scalars or coordinates are supported."
```

gonuke (changes requested):
A few more thoughts on comments for clarity

PR #249, opened 2025-07-31
Update Dockerfile and Hotfix Dependency Installation

connoramoreno:
Copy `pyproject.toml` in Dockerfile

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,3 +41,6 @@ jobs:
 
       - name: Test running as executable module
         run: python -m parastell -h
+
+      - name: Test Python import
+        run: python -c "import parastell"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,7 @@ jobs:
           environment-file: environment.yml
 
       - name: Install with pip
-        run: pip install .[develop]
+        run: pip install --no-deps .
       
       - name: Test building wheels
         run: python -m build --sdist --wheel
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,6 @@ jobs:
         run: |
           . /opt/etc/bashrc
           sed -e "s/@SERVER@/${rlmSERVER}/" -e "s/@PASSWORD@/${rlmPASSWD}/" /opt/Coreform-Cubit-2024.8/bin/licenses/rlmcloud.in > /opt/Coreform-Cubit-2024.8/bin/licenses/rlmcloud.lic
-          export PYTHONPATH=${PYTHONPATH}:`pwd`
           cd tests
           pytest -v .
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Regular CI testing for parastell
+name: Regular CI testing for ParaStell
 
 on:
   # allows us to run workflows manually
@@ -24,7 +24,7 @@ jobs:
         run: |
           . /opt/etc/bashrc
           sed -e "s/@SERVER@/${rlmSERVER}/" -e "s/@PASSWORD@/${rlmPASSWD}/" /opt/Coreform-Cubit-2024.8/bin/licenses/rlmcloud.in > /opt/Coreform-Cubit-2024.8/bin/licenses/rlmcloud.lic
-          cd tests
+          cd /opt/parastell/tests
           pytest -v .
         env:
           rlmSERVER: ${{ secrets.RLMSERVER }}
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -47,6 +47,8 @@ jobs:
       - name: Build and push ParaStell Docker image
         id: build-parastell
         uses: docker/build-push-action@v5
+        run:
+          pip install .[develop]
         with:
           no-cache: ${{ github.event.inputs.use_cache == 'false' }}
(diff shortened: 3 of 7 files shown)


gonuke:
> The in-progress check has been stuck on the same step for several hours now...

It was stuck on the `pip install` maybe it either couldn't find the code (expecting it to be local) or couldn't access the repo?

connoramoreno:
I think we either need to run `pip install .[develop]` in each test workflow (since it's not part of the Docker image) or change the Docker image to include a full ParaStell install. At least, those are the solutions I can think of. The latter seems more reasonable to me.

connoramoreno:
I noticed in a previous run that the log does not update after about 66% through the Cubit download (too many lines?). Assuming the log is showing everything once the run has been canceled, it seems like it's getting stuck on the step
```#19 1.698   Getting requirements to build wheel: started```

connoramoreno:
Good catch, I missed that

gonuke:
This might be the cleanest answer: https://stackoverflow.com/a/71163579

connoramoreno:
I believe I've diagnosed the issue. It's not coming from Cubit after all, but Gmsh. Our new installation has two primary steps:

1. Conda install using `environment.yml`
2. pip install using `pyproject.toml`.

The Conda install of Gmsh actually installs three things:

1. `gmsh` - compiled Gmsh binary
2. Gmsh libraries
3. `python-gmsh` - Python wrapper for Gmsh API, compiled against above-mentioned Gmsh libraries

When the Conda install is followed by the pip install, pip finds the version of Gmsh on PyPI and overwrites the Conda-installed `gmsh` package, but `python-gmsh` remains in the environment. For some reason this isn't done for, e.g., CadQuery or CAD-to-DAGMC, which are also on PyPI.

Anyways, at this point the Gmsh Python wrapper no longer matches the binary/libraries, which I think is causing the segmentation fault.

The only real solution I can think of is to remove the listed dependencies from `pyproject.toml` and revert to our previous version of `environment.yml` that included the pip-installed packages, but this is of course not following best practices. Your thoughts @gonuke @MicahGale?

connoramoreno:
In my local testing, I pinned the minor version of Gmsh in both `environment.yml` and `pyproject.toml`, and although `gmsh` is still overwritten with the PyPI version, tests now pass. That said, I'm not sure that's a solution I'm comfortable with. I'd rather have `gmsh` and `python-gmsh` come from the same manager.

gonuke:
> I believe I've diagnosed the issue. It's not coming from Cubit after all, but Gmsh. Our new installation has two primary steps:
> 
> 1. Conda install using `environment.yml`
> 2. pip install using `pyproject.toml`.
> 
> The Conda install of Gmsh installs three things:
> 
> 1. `gmsh` - compiled Gmsh binary
> 2. Gmsh libraries
> 3. `python-gmsh` - Python wrapper for Gmsh API, compiled against above-mentioned Gmsh libraries
> 
> When the Conda install is followed by the pip install, pip finds the version of Gmsh on PyPI and overwrites the Conda-installed `gmsh` package, but `python-gmsh` remains in the environment. For some reason this isn't done for, e.g., CadQuery or CAD-to-DAGMC, which are also on PyPI.
> 
> Anyways, at this point the Gmsh Python wrapper no longer matches the binary/libraries, which I think is causing the segmentation fault.
> 
> The only real solution I can think of is to remove the listed dependencies from `pyproject.toml` and revert to our previous version of `environment.yml` that included the pip-installed packages, but this is of course not following best practices. Your thoughts @gonuke @MicahGale?

Since I don't see it listed explicitly, I assume that `gmsh` is a dependency of something in the `environment.yml` file and also in the `pyproject.toml` file.  Since we need `pygmsh` from conda, can we figure out what is causing `gmsh` to get pulled from PyPI and just move that to the `environment.yml` file?  Or is there a `pip` setting that can avoid overwriting existing packages?

connoramoreno:
Sorry, I realize now that I didn't push my local version that has Gmsh listed explicitly. It's a dependency of CAD-to-DAGMC but I think it makes sense to list it directly anyways since we use it directly. CAD-to-DAGMC is listed in both `environment.yml` and `pyproject.toml`.

I spent a while looking into it and it doesn't seem like there's anything I can add to `pyproject.toml` to prevent `pip` from overwriting existing packages.

MicahGale:
> I spent a while looking into it and it doesn't seem like there's anything I can add to pyproject.toml to prevent pip from overwriting existing packages.

I think this is more an issue that `gmsh` isn't being properly installed by `conda` then.

connoramoreno:
Just pushed my changes from last week - it's my most up-to-date version but still broken

connoramoreno:
> Should we do `pytest` outside of docker as well?

Is there a way to do that in our workflows?

In case it wasn't clear, I tested the installation process locally and ran into the same Gmsh issue that our CI image is having.

MicahGale:
I'd like to see that the installed version passes at least some test. Could you run `pytest` in `build.yml`. Probably have it skip the cubit tests though.

connoramoreno:
> I'd like to see that the installed version passes at least some test. Could you run `pytest` in `build.yml`. Probably have it skip the cubit tests though.

The installed version is passing the test suite in the `Test CI Image` part of the `Build & publish Docker image for ParaStell CI` workflow. The `Install Dependencies` part of that workflow builds its environment based on this branch's `environment.yml` and `pyproject.toml`

MicahGale (approved):
This seems to be the best we can do for now, but I think it will be important to get `parastell` onto conda-forge ASAP to hide this shenanigans from users.

PR #252, opened 2025-08-07
Update logo, example image, and README

connoramoreno:
Updates the logo with a new color palette, the example image to reflect that new palette, and the README


connoramoreno:
Not sure why tests aren't passing but I don't think it has to do with this PR

pyproject.toml
```diff
@@ -74,3 +74,6 @@ build-backend = "setuptools.build_meta"
 [tool.black]
 line-length = 79
 target-version = ['py311', 'py312', 'py313']
+
+[tool.setuptools.packages.find]
+exclude = ["images"]
```
gonuke, line 79: If we wanted the images to be installed (not important), is there a way to do it?

gonuke (commented):
Just one question about the `images` directory...

pyproject.toml
connoramoreno: My understanding is that `pip` couldn't tell which directory (`images` or `parastell`) has the source code, so this directory is excluded from any wheel building

gonuke: Is that because of the specific name `images` or just because it was some directory that is not some other standard name?

MicahGale: Yes @gonuke pip gets confused by multiple non-standard folders. It only knows to exlcude `test*`, `doc`, `build`, `dist`, etc.

I think images shouldn't be included with a package anyways.

MicahGale (approved):
I like these pretty colors.

PR #254, opened 2025-08-28
Fix assignment of volume IDs to constructed magnet solids

connoramoreno:
For magnets with casing constructed from filaments, volume IDs were being incorrectly assigned such that IDs were being repeated. This branch fixes that issue.

### DIFF
--- a/tests/test_parastell.py
+++ b/tests/test_parastell.py
@@ -44,8 +44,6 @@ def check_surfaces_and_volumes(filename, num_surfaces_exp, num_volumes_exp):
     """
     dagmc_model = pydagmc.Model(str(Path(filename).with_suffix(".h5m")))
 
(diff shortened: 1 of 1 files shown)


connoramoreno:
I suspect we are finally running into the Cubit license issue

connoramoreno:
No, I'll write a new test

connoramoreno:
@gonuke I've expanded testing and tests are passing locally - Looks like the license server is having some difficulty at the moment

connoramoreno:
Looking at our Coreform account, it looks like I'm using all our license seats, which is likely why the test is failing as it is, but I don't actually have any Cubit processes running (I don't think)... Do the containers stay running after a workflow has failed?

gonuke (approved):
This looks OK to me now.  Thanks @connoramoreno

PR #255, opened 2025-08-28
Update to new Cubit license system

connoramoreno:
Updates ParaStell's CI workflow in accordance with Coreform's new system for Cubit's license server. As a result, also updates the version of Cubit used for CI to 2025.8. Makes use of new GitHub secrets.


connoramoreno:
@gonuke Correct!

PR #256, opened 2025-09-02
Log out of license server after CI tests

connoramoreno:
Adds a command in the CI workflow files to log out of Cubit's RLM server after tests are run


gonuke (approved):
Thanks for confirming this @connoramoreno

PR #257, opened 2025-09-11
Change handling of 360-degree geometries for magnets

connoramoreno:
This branch modifies the magnet creation workflow such that when `toroidal_extent = 360.0`, the intersection operation of each magnet coil with `toroidal_domain` does not occur. This does not cause any issues, but is an unnecessary operation.


gonuke (approved):
LGTM - thanks @connoramoreno

PR #259, opened 2025-09-19
Add radial component to closeness evaluation

connoramoreno:
Not to make the volume ID assignment algorithm more of a beast, but I realized in testing that the condition for evaluating the closeness of solids was insufficient for geometries that are cut at the ends of the toroidal domain of the model. In such cases, the toroidal angles of the centers of mass of the cut solids tend to be quite close to each other. To solve this issue, I added a radial component to the evaluation.


gonuke (changes requested):
some updated variable names

PR #262, opened 2025-11-04
Separate particle processing into statistical batches in NWL calculation

connoramoreno:
Converts the NWL calculation's batching process into a statistical batching routine. Previously, the batching process was used only to alleviate memory burden and batch counts were combined into a single binning calculation. Now, batch counts are binned separately. This enables statistical batching for the NWL calculation, for the purpose of estimating variance.

Since the source is sampled randomly, it is assumed that each batch is independent, even though particle crossings come from a single OpenMC batch.

### DIFF
--- a/parastell/nwl_utils.py
+++ b/parastell/nwl_utils.py
@@ -332,7 +332,7 @@ def compute_nwl(
 
     toroidal_angles = []
     poloidal_angles = []
-    all_nwl = []
+    nwl_all_batches = []
 
     batch_size = math.ceil(num_particles / num_batches)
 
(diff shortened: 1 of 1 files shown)


connoramoreno:
Right, I hadn't actually worked out the math before... is this still worthwhile?

I also realized that the final conversion from counts to NWL (in this version) is incorrect since we're not using batch-wise particle numbers.

tokamaster:
I think having a 'native' way to estimate the variance is a good idea, specially if it's not a huge headache to implement. I can already calculate the variances by doing external batching, but getting it handled directly could make it easier for the user.

My results indicate that the variance increases in areas with low NWL, which I think intuitively makes sense, as fewer particles are hitting those spots. For tokamaks, this might not be super interesting because their NWL is generally uniform toroidally, but for stellarators, finding those higher-variance spots could be a valuable clue about the impact of shaping and geometry...

gonuke:
If you want to expand this to estimate variance, I think it makes sense, and might not be too hard

parastell/nwl_utils.py
```diff
@@ -400,9 +416,17 @@ def compute_nwl(
                 corners
             )
 
-    nwl_mat = nwl_mat / area_mat
+    # Normalize NWL statistics to area
+    nwl_avg /= area_mat
+    nwl_std_dev /= area_mat
```
gonuke: This math may not be right for the std_dev?

connoramoreno: Your math looks good to me. As demonstrated by your proof, the [standard deviation scales directly with an applied constant](https://en.wikipedia.org/wiki/Standard_deviation#Identities_and_mathematical_properties)

connoramoreno: If you'd prefer, I can divide the original data set by the area and then calculate the mean and standard deviation

gonuke (approved):
Thanks @connoramoreno - this will be a useful addition

PR #270, opened 2026-04-21
Add NWL export to ASCII text format

EgorVe:
A small function to export neutron wall loading results from compute_nwl() as a human readable 4-column ASCII file:
- Toroidal angle [deg]
- Poloidal angle [deg] 
- NWL [MW/m²]
- Area [m²]

### DIFF
--- a/parastell/nwl_utils.py
+++ b/parastell/nwl_utils.py
@@ -470,3 +470,33 @@ def plot_nwl(
 
     export_path = Path(filename).with_suffix(".png")
     fig.savefig(export_path)
+
+
+
+def save_nwl_ascii(
+    nwl_mat,
+    area_mat,
+    toroidal_centroids,
+    poloidal_centroids,
+    filename="nwl",
+):
+    """Saves NWL as four-column ASCII:
+    toroidal_bin_deg, poloidal_bin_deg, nwl_value, area_value.
+    """
+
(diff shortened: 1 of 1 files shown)


gonuke:
Thanks for this contribution @EgorVe - seems like a great idea.  Is there a reason you chose space-delimited columns instead of CSV?  It seems that we could take advantage of built-in CSV capability for more robust but still human-readable formatting.

EgorVe:
@gonuke - I used that format because one of our students specifically asked for it. But yes, I agree CSV is more robust and widely usable option, so I’ll update the function accordingly.
Not totally sure what do you mean by “built-in CSV capability” though, could you please clarify?

gonuke:
> @gonuke - I used that format because one of our students specifically asked for it. But yes, I agree CSV is more robust and widely usable option, so I’ll update the function accordingly. Not totally sure what do you mean by “built-in CSV capability” though, could you please clarify?

"built-in CSV capability" - Python has methods to conveniently read and write CSV, that's all meant, including from Numpy and Pandas

EgorVe (commented):
I’ve updated the NWL text file export to CSV format. The new version keeps the same data fields, but writes them in a 4-column CSV layout using NumPy CSV functionality.

gonuke (approved):
This looks good to me - thanks @EgorVe

